PR #4, opened 2021-06-10
Persistent storage

StewartJingga:


recipes/store.go
```diff
@@ -6,46 +6,3 @@ type Store interface {
 	GetByName(string) (domain.Recipe, error)
```
ravisuhag, line 6: We can move this package to root and call it storage. As it used by recipe but not directly dependent.

StewartJingga: done! also changed a few things to accommodate it.

ravisuhag: Does it make sense to name it storage? as only one backend will be enabled at one time.

StewartJingga: Tbh I don't have any preferences between store and storage as long as it is immediately understandable and not ambiguous. Also it should be easy change, so are you talking only about the package naming or the whole thing (e.g. `RecipeStorage`)?

ravisuhag: Ok. Let's keep it store then.

PR #6, opened 2021-06-16
Build secrets system

StewartJingga:


stores/secret_store.go
```diff
@@ -0,0 +1,75 @@
...
+
+	err = json.NewDecoder(r).Decode(&secret)
+	if err != nil {
+		return secret, err
+	}
+
+	return
+}
+
+func (s *secretStore) Upsert(secret secrets.Secret) (err error) {
```
rahmatrhd, line 50: Can we encrypt the secrets before storing them? Similar with when we are only storing user's encrypted password for the privacy concern.

PR #87, opened 2021-06-26
feat: mongodb extractor

GrayFlash:
issue #81


extractors/mongodb/extractor.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return
+	}
+	uri := "mongodb://" + config.UserID + ":" + config.Password + "@" + config.Host
+	clientOptions := options.Client().ApplyURI(uri)
+	result, err = ListCollections(clientOptions)
+	if err != nil {
+		return
+	}
+	fmt.Println(result)
```
StewartJingga, line 44: don't forget to remove this 😄

extractors/mongodb/extractor.go
```diff
@@ -0,0 +1,118 @@
...
+	return result, err
+}
+
+func ListCollections(clientOptions *options.ClientOptions) (result []Result, err error) {
```
StewartJingga, line 48: should we change this to private function?
wdyt?

extractors/mongodb/extractor.go
```diff
@@ -0,0 +1,118 @@
...
+			row.CollectionName = collection
+			row.DatabaseName = db_name
+			row.Indexes = ListIndexes(clientOptions, collection, db_name)
+			result = append(result, row)
+		}
+	}
+	return result, err
+}
+
+func ListIndexes(clientOptions *options.ClientOptions, collection string, db_name string) (results []bson.D) {
```
StewartJingga, line 80: this too, should we change it to private?

extractors/mongodb/extractor_test.go
```diff
@@ -0,0 +1,178 @@
...
+type Reach struct {
+	Views    string `bson:"views,omitempty"`
+	Likes    string `bson:"likes,omitempty"`
+	Comments string `bson:"comments,omitempty"`
+}
+
+func TestExtract(t *testing.T) {
+	t.Run("should return error if no user_id in config", func(t *testing.T) {
+		extractor := new(mongodb.Extractor)
+		_, err := extractor.Extract(map[string]interface{}{})
```
StewartJingga, line 53: let's put password field here to make sure the error is not because of missing password.

StewartJingga: also host

extractors/mongodb/extractor_test.go
```diff
@@ -0,0 +1,178 @@
...
+		extractor := new(mongodb.Extractor)
+		_, err := extractor.Extract(map[string]interface{}{
+			"user_id":  "Gaurav_Ubuntu",
+			"password": "abcd",
+		})
+
+		assert.NotNil(t, err)
+	})
+
+	t.Run("should return error when testing without mongo running", func(t *testing.T) {
```
StewartJingga, line 77: after some thoughts, I think we can remove this test as this test is to test testing (hope it makes sense 😅 ).
wdyt?

extractors/mongodb/extractor_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	defer cur.Close(ctx)
+	var connectionsDB []Connection
+	if err = cur.All(ctx, &connectionsDB); err != nil {
+		return err
+	}
+	return
+}
+
+func InsertReach(ctx context.Context, client *mongo.Client) (err error) {
```
StewartJingga, line 162: all these functions could be change to private if we are not exposing it elsewhere.

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '3'
+services:
+  mongo:
+    image: mongo
+    environment:
+      - MONGO_INITDB_ROOT_USERNAME=user
+      - MONGO_INITDB_ROOT_PASSWORD=abcd
+    restart: on-failure
+    ports:
+      - 27017:27017
```
StewartJingga, line 10: let's put newline here.

extractors/mongodb/extractor.go
GrayFlash: This is the final result, and it needs to be shown on console right??

extractors/mongodb/extractor.go
GrayFlash: ok, so I make these methods private, since we are not using them outside the packages.

extractors/mongodb/extractor.go
StewartJingga: oh no, we show it on console using sink, extractor will only fetch metadata.

extractors/mongodb/extractor_test.go
GrayFlash: okay!

extractors/mongodb/extractor_test.go
GrayFlash: Yeah, but in a test set-up we need to test if for a given config we can connect to a mongo server right?

StewartJingga: oh in that case, instead of testing this, we could test if `Extract()` returns the expected data instead.

PR #108, opened 2021-06-28
feat: setup telemetry

StewartJingga:
Enable sending metrics via statsd


recipes/monitor.go
```diff
@@ -0,0 +1,9 @@
+package recipes
+
+type Monitor interface {
+	RecordRun(recipe Recipe, durationInMs int, success bool)
+}
+
+type defaultMonitor struct{}
```
rahmatrhd, line 7: this `defaultMonitor` gonna be used if the statsd isn't enabled, right? If yes, what is the consideration to create this 'empty' defaultMonitor instead of using condition such as `if statsdEnabled { monitor.RecordRun(...) }`?

StewartJingga: there are a few reasons:
- statsd will not be known to runner, runner can only check if monitor is nil or not.
- I am using [null object pattern](https://sourcemaking.com/design_patterns/null_object) to reduce if statement `(if monitor == nil)` in `runner.Run()` so it can focus on it's core logic.

rahmatrhd: TIL null object pattern. Got it 👍

rahmatrhd (approved):
LGTM

PR #116, opened 2021-07-02
feat: mysql extractor

GrayFlash:


extractors/mysql/extractor.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+type Extractor struct{}
```
StewartJingga, line 11: if you could try to put this Extractor struct definition near its methods.

extractors/mysql/extractor.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return
+	}
+	err = e.validateConfig(config)
+	if err != nil {
+		return
+	}
+	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s)/", config.UserID, config.Password, config.Host))
+	if err != nil {
+		fmt.Println(err)
```
StewartJingga, line 37: dont forget to remove printing to console

extractors/mysql/extractor.go
```diff
@@ -0,0 +1,145 @@
...
+		var database string
+		res.Scan(&database)
+		if checkNotDefaultDatabase(database) {
+			result, _ = tableInfo(database, result, db)
+		}
+	}
+	return
+}
+
+func tableInfo(dbName string, result []map[string]interface{}, db *sql.DB) (_ []map[string]interface{}, err error) {
```
StewartJingga, line 59: we could change the method name to  represent what it does clearly.
right now it receives a list (result) and populate it with another table information.

so probably we can change it to `populateWithTables`.
wdyt?

extractors/mysql/extractor.go
```diff
@@ -0,0 +1,145 @@
...
+		row["field_desc"] = fieldDesc
+		row["data_type"] = dataType
+		row["is_nullable"] = isNull
+		row["length"] = length
+		result = append(result, row)
+	}
+	return result, nil
+}
+
+func (e *Extractor) getConfig(configMap map[string]interface{}) (config Config, err error) {
```
StewartJingga, line 120: make sure all struct methods are not split.
currently `getDatabases`, `tableInfo`, and `fieldInfo` are in between these struct methods.

either you change those functions to struct methods or move those function to the bottom.
Personally I would suggest changing those to struct methods instead.

extractors/mysql/extractor_test.go
```diff
@@ -0,0 +1,212 @@
...
+	tableQuery := "CREATE TABLE "
+	err = createTable(tableQuery, table1, var1, db)
+	if err != nil {
+		return
+	}
+	err = createTable(tableQuery, table2, var2, db)
+	if err != nil {
+		return
+	}
+	_, err = db.Exec("CREATE USER IF NOT EXISTS 'user'@'%' IDENTIFIED BY 'pass';")
```
StewartJingga, line 161: we could use our `user` and `pass` variables to make sure everything is in sync.

extractors/mysql/extractor.go
GrayFlash: I think ```getTablesInfo``` will be a better name, actually it takes in a dbName, and appends the information of all the tables in that db, but since we are dealing with db's and it performs a ```SELECT``` operation and not ```INSERT``` I think populateWithTables might be confusing??

StewartJingga (approved):

PR #118, opened 2021-07-05
feat: postgres extractor

GrayFlash:


extractors/postgres/extractor_test.go
```diff
@@ -0,0 +1,236 @@
...
+	db.Close()
+	db, err = sql.Open("postgres", fmt.Sprintf("postgres://%s:%s@localhost:5432/%s?sslmode=disable", user, pass, testDB))
+	if err != nil {
+		return
+	}
+	table1 := "applicant"
+	var1 := "(applicant_id int, last_name varchar(255), first_name varchar(255))"
+	table2 := "jobs"
+	var2 := "(job_id int, job varchar(255), department varchar(255))"
+	tableQuery := "CREATE TABLE "
```
StewartJingga, line 179: CMIIW but I think we could move this to `createTable` and since it will only be used in that function.

StewartJingga: also for readability maybe we could do this
```
err = createTable(db, "applicant",  columnsQuery)  // columnsQuery is var1
```

.github/workflows/test.yml
```diff
@@ -20,9 +20,8 @@ jobs:
           MYSQL_ALLOW_EMPTY_PASSWORD: true
       postgres:
         image: postgres:13.3
-        restart: always
         ports:
-          - "5432:5432"
+          - 5432:5432
         env:
           - POSTGRES_USER: root
           - POSTGRES_PASSWORD: pass
```
StewartJingga: does postgres have a way to use default root user name and default password?

GrayFlash: Actually it was throwing error without these, and other option was to allow all host connections to use the db without authentication ```POSTGRES_HOST_AUTH_METHOD=trust```, but I think this is not recommended.

GrayFlash: Btw, I have used this as a root to create another temporary user for that particular db. So, we are not accessing the extractor as a root, and I think it should be fine.

StewartJingga (approved):

PR #121, opened 2021-07-08
feat!: change store to factory

StewartJingga:


ravisuhag:
@StewartJingga How about other way around? Moving extractor implementation to root and other to pkg?

StewartJingga:
@ravisuhag oh actually they are not plugins. By themselves, extractors, sink and processors and built-in and are the main packages.

Plugin is a package to handle custom code via rpc which at the moment only supports processor.

extractors/factory.go
```diff
@@ -0,0 +1,26 @@
+package extractors
+
+type FactoryFn func() Extractor
```
GrayFlash, line 3: FactoryFn is an abbreviation of Factory function right??

StewartJingga: yes 👍

plugins/external/discover.go
```diff
@@ -4,7 +4,7 @@ import (
 	"os"
```
ravisuhag, line 4: @StewartJingga How about moving this external package to core and naming it plugin. Since this package is about providing core functionality of supporting a plugin system in meteor but it is not really a plugin. Is that correct understanding or am I missing something?

StewartJingga: im thinking core package is something that is fully consist of core system and logic without any interface contexts and super minimum libraries (e.g. http, rpc, cli, statsd etc).

if you check current packages inside `core` package, they are mostly plain native golang struct with minimum library or package. (it only uses yaml library)

right now external plugin has a heavy rpc context and `go-plugin` library and putting them inside core package would tear down the scope of core package.

the other reason I put external plugins to plugins folder is because I'm guessing when people are looking at our code, whenever they are looking for external plugins that would be the first folder they go into.

wdyt? @ravisuhag

StewartJingga: Also atm, I do not see plugins folder as the a mere folder to store all the plugins, I see it as a plugins package which contains means to fetch our internal plugins as well as external plugins.

ravisuhag: sounds good!

PR #127, opened 2021-07-21
feat: add logger and extractor docs

StewartJingga:


cmd/rundir.go
```diff
@@ -13,19 +12,19 @@ func rundir(dirPath string) {
...
 
 	reader := recipe.NewReader()
 	recipeList, err := reader.ReadDir(dirPath)
 	if err != nil {
 		panic(err)
 	}
-	runner, cleanFn := initRunner(c)
+	runner, cleanFn := initRunner(c, log)
 	defer cleanFn()
 	faileds, err := runner.RunMultiple(recipeList)
```
GrayFlash, line 27: ```faileds``` can be renamed to ```failed_recipes``` or ```failures```, wdyt??

StewartJingga: yup definitely can be changed.
how about to `failedRecipes` for following convention?

GrayFlash: Yup seems better

GrayFlash:
@StewartJingga seems fine to me!!

PR #133, opened 2021-07-22
feat: add command to scaffold extractor

scortier:
#### Command : `meteor rungen --type <type_of_extractor> --extractor <name_of_extractor>`


ravisuhag:
Command interface could be more intuitive.
`$ meteor generate extractor --type=topic --name="postgres"`

scortier:
> Command interface could be more intuitive.
> `$ meteor generate extractor --type=topic --name="postgres"` or
> `$ meteor new extractor --type=topic --name="postgres"`

Sure ,will update it

ishanarya0:
Right, this would be better. We'll modify it.

ravisuhag:
Also, this command would require to have a template module that will have code generation in place to generate sample code for an extractor.

ravisuhag (changes requested):
Add template code generation and update comamnd

StewartJingga:
> Command interface could be more intuitive.
> $ meteor generate extractor --type=topic --name="postgres" or
> $ meteor new extractor --type=topic --name="postgres"

`$meteor generate extractor --type=topic --name=postgres` looks good

StewartJingga (changes requested):
Move generation logic to its own package.

The generation logic should be its own package, `cmd` package should only focus on gathering inputs and building dependencies.

ishanarya0:
Added template module and file generation

WIP: Writing template content to generated files

PR #137, opened 2021-07-22
feat: support streamable payloads

kushsharma:
- This will enable a streamable payload that can drastically improve the memory overhead for each runner
- Single interface for all extracts
- No abstraction leak in sinks, earlier implementation needed sink to know in advance what the extractor is writing(datatype) now we are adding limitations to sinks via Codecs, so an extractor needs to implement different codecs to support different types of sinks if needed. (I still need to confirm if this works for proto)
- No need to register protobuf models for each extractor, only extractors who want to support sinks that need Proto codec will require proto models to be registered. For example, I have added a `date` extractor to demonstrate a JSON codec
- Runtime preprocessors are broken at the moment
- Existing extractors need to be refactored to use new implementation


core/interface.go
```diff
@@ -0,0 +1,21 @@
+package core
+
+type Extractor interface {
```
ravisuhag, line 3: Let's extend this interface to also include a description of the extractor. 
`Description ` as well as `Name` The same name can then be used in the recipe when defining source or sink etc.

ravisuhag (commented):
- We can extract the common package code from cmd to a runner.
- Sub packages in core for extractor, sink and processors can be composed in one package as `registry/catalog` or even core for that matter instead of three sub packages. 
- With that recipe from core can be extracted out as root package. 

Proposal. 
```
runner
registry
recipe
plugins
```

core/interface.go
kushsharma: Where are we going to print this description?

cmd/common.go
```diff
@@ -9,47 +9,31 @@ import (
...
 		killPluginsFn()
 	}
 	return
 }
-func initExtractorFactory(logger plugins.Logger) *extractor.Factory {
-	factory := extractor.NewFactory()
-	extractorPlugins.PopulateFactory(factory, logger)
-	return factory
-}
 func initProcessorFactory() (*processor.Factory, func()) {
```
ravisuhag, line 28: Can this be encapsulated within processor.Catalog to make programming interfaces consistent with sink and extractor.

kushsharma: oh yeah, this is a casualty of merge conflict 😭

kushsharma: Nevermind, this was done on purpose actually. I left it unchanged as I need to refactor how runtime plugins are being instantiated. Because runtime plugins are currently broken so didn't fix it. I'll do it later once rest of the implementation looks concrete. For now, I'll just disable runtime plugins completely?

ravisuhag: @StewartJingga ^^

kushsharma: I have disabled runtime plugins in this PR for now.

kushsharma:
> @kushsharma @StewartJingga
> 
> * We can extract the common package code from cmd to a runner.
> * Sub packages in core for extractor, sink and processors can be composed in one package as `registry/catalog` or even core for that matter instead of three sub packages.
> * With that recipe from core can be extracted out as root package.
> 
> Proposal.
> 
> ```
> runner
> registry
> recipe
> plugins
> ```

Sounds good to me. I didn't reorganize the code to avoid too much re-write. Will do this ^^

plugins/extractors/date/date.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (p *Payload) ToProto() ([]byte, error) {
+	protoC := &facets.Custom{CustomProperties: map[string]string{
+		"Value": p.Value,
+	}}
+	return proto.Marshal(protoC)
+}
+
+func (d *Extractor) Extract(ctx context.Context, config map[string]interface{}, out chan<- interface{}) (err error) {
```
ravisuhag, line 39: Using go channels approach will increase complexity for end-users to write extractors. Ideally, we want the extractor to be very dumb and easy to implement.

plugins/extractors/date/date.go
```diff
@@ -0,0 +1,81 @@
...
+	Value string
+}
+
+type Payloads []Payload
+
+func (p *Payload) ToJSON() ([]byte, error) {
+	return json.Marshal(p)
+}
+
+func (p *Payload) ToProto() ([]byte, error) {
```
ravisuhag, line 32: For any user to writing a metadata extractor, we want them to extract in a proper schema. e.g. dataset or a user. we want this data to be mapped to a model. Wire format and sink format could be anything JSON, text, proto, or anything but ideally, extractor schema should be consistent with one of the models. So with that in mind, does it make sense to stick that users use one of the proto-model to model their metadata and codec can be a central module which allows sinks to process and dump in whatever format they would like to choose?

plugins/extractors/date/date.go
StewartJingga: IMHO we could probably abstract it but even then not sure how much simpler it would be. 🤔

plugins/extractors/date/date.go
StewartJingga: Ok so I would propose that serialization should not be the concern of extractors and processors, it should be the concern for sinks. We could have a serializer module that would do the heavy lifting for serialization, so even an individual sink should not care about it and just the module.

What to serialise data into should not be decided by the extractor, it should be up to the user. For example it would be totally possible to use kafka sink and serialise it into json format, or maybe sinking it to stdout and serialize it to protobuf (not sure why users would do that but again we don't care).

Since all the serialization logic would be on the serializer module, even if we somehow have to do 10-20 switch cases, it is still being contained there and other sinks can just use it without having to rewrite the same thing.

Building proto key can also be handled in that same module, and since these metadata models (Table, Dashboard, etc) are maintained by us, it would make sense as well that we decide what would be the proto key. I dont see a reason to make it customisable for now.

Sample serializer usage in recipe
```
source:
  type: bigquery
  config: {}
sinks:
 - name: kafka
    serializer: json
    config:
      brokers: localhost:9092,localhost:9093
      topic: bigquerytable_metadata
 - name: console
    serializer: protobuf
```

what do you guys think? @ravisuhag @kushsharma

plugins/extractors/date/date.go
StewartJingga: @ravisuhag  btw this would become what we are avoiding in the first place right, allowing them to output dynamic data.

plugins/extractors/date/date.go
```diff
@@ -0,0 +1,81 @@
...
+	time.Sleep(500 * time.Millisecond)
+	// ...
+
+	return &Payload{
+		Value: fmt.Sprintf("Timestamp: %s", time.Now().String()),
+	}, nil
+}
+
+func init () {
+	if err := extractor.Catalog.Register("date", &Extractor{
```
StewartJingga, line 75: this works fine for logger since most times logger is stored as package variable.

but what if we want to inject monitoring dependency or others here? for example, there is statsd client being instantiated in cmd, how do we pass it to this extractor? do we store that client as package variable too?

StewartJingga:
> @kushsharma @StewartJingga
> 
> * We can extract the common package code from cmd to a runner.
> * Sub packages in core for extractor, sink and processors can be composed in one package as `registry/catalog` or even core for that matter instead of three sub packages.
> * With that recipe from core can be extracted out as root package.
> 
> Proposal.
> 
> ```
> runner
> registry
> recipe
> plugins
> ```

Agree with moving runner to its own package and recipe to core 🆗 

But Im still seeing not much benefits about joining extractors, sinks and processors. As far as im concerned, no cleanliness or dependencies management are being improved here 🤔, this is mostly about structuring packages right?  And my idea of them having their own packages is so that any logic related to each of them are stored within their own package. Having them all one package may also lead to harder variable naming due to possible conflict and mixed contexts.

Also package naming is pretty important for readability in go as separating would allow me to do `extractor.NewFactory()` instead of `core.NewExtractorFactory()`.

wdyt? @ravisuhag @kushsharma

plugins/extractors/date/date.go
ravisuhag: Each sink will do a hard check of the type if needed. So we can keep it this way.

PR #148, opened 2021-07-26
refactor: cleanup postgres extractor

ravisuhag:
- Refactor extractor to allow exclude instead of hardcoded databases. 
- Fix bug to allow fetching table metadata from multiple databases.
- Refactor tests to make them DRY. 
- Refactor extractor to stream table one at a time.


plugins/extractors/postgres/postgres.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/odpf/meteor/proto/odpf/meta/facets"
+	"github.com/odpf/meteor/utils"
+)
+
+type Config struct {
+	UserID   string `mapstructure:"user_id" validate:"required"`
+	Password string `mapstructure:"password" validate:"required"`
+	Host     string `mapstructure:"host" validate:"required"`
+	Database string `mapstructure:"database" default:"postgres"`
+	Exclude  string `mapstructure:"exclude" default:"information_schema,root,postgres"`
```
GrayFlash, line 22: Do we mean that we are allowing user to pass which databases to exclude from extraction and set those hardcoded values of default db as default here in case they don't add anything??

GrayFlash: Also, user will have to search these default db's himself to exclude them, If at all he needs to exclude some other db's as well. Since if they pass Exclude value as ```db1, db2```. Extractor will only exclude these 2 and not the ```information_schema,root,postgres``` right?

GrayFlash: Btw, the test and setup looks much cleaner, will implement those with others!!

plugins/extractors/postgres/postgres.go
```diff
@@ -0,0 +1,201 @@
...
+func (e *Extractor) getDatabases(cfg Config, db *sql.DB) (list []string, err error) {
+	res, err := db.Query("SELECT datname FROM pg_database WHERE datistemplate = false;")
+	if err != nil {
+		return nil, err
+	}
+
+	for res.Next() {
+		var database string
+		res.Scan(&database)
+		if exclude(cfg, database) {
```
StewartJingga, line 90: this exclude will split a string and transform it into array for every table in every database.
should transform it once and store the array to the extractor struct?

also I was thinking we could probably use map instead of array for quicker check since Exclude is now configurable. (edge cases where users excludes lots of table)

GrayFlash:
Also any reason for renaming the ```extractor.go``` to ```postgres.go```?? I think our package name is ```postgres``` already so I thought ```extractor.go``` was fine seeing the intent of that code.

plugins/extractors/postgres/postgres.go
GrayFlash: Yeah, so we can split the Exclude string once and map the values? That will be a one time operation, so it should be efficient?

plugins/extractors/postgres/postgres.go
ravisuhag: @GrayFlash Yes, this is to allow people to exclude databases. 
Make sense. We can take this into account to add these to the list.

plugins/extractors/postgres/postgres.go
ravisuhag: > Also any reason for renaming the `extractor.go` to `postgres.go`?? I think our package name is `postgres` already so I thought `extractor.go` was fine seeing the intent of that code.

File searching in the editor becomes very tricky with all files named extractors. And general Go practice is also to use the main file of the package as the same name.

ravisuhag:
File searching in the editor becomes very tricky with all files named extractors. And general Go practice is also to use the main file of the package as the same name.

plugins/extractors/postgres/postgres.go
StewartJingga: @GrayFlash yes, transforming the exclude string to a map (e.g. map[string]bool) will be a once for one extraction.
Also when checking each of table if it is excluded will be constant compared to if it is an array (we have to loop through all the items in the array for almost every table).

StewartJingga (approved):

PR #149, opened 2021-07-27
feat: modify extractors to stream data

StewartJingga:
Changes:
1. Update extractor, processor and sink factory to return a new instance everytime
2. Update extractors:
-  csv
-  elasticsearch
-  gcs
-  grafana
-  kafka
-  mongodb
-  mssql
-  mysql


core/extractor/factory.go
```diff
@@ -10,26 +10,26 @@ var (
 )
 
 type Factory struct {
-	fnStore map[string]core.Extractor
+	fnStore map[string]func() core.Extractor
```
kushsharma, line 13: any reason we are using these functions here instead of directly interface?

StewartJingga: ok so basically I want this factory to return a new instance of extractor for each run.

reasons are:
1. this would allow extractor to utilise its private fields.
2. and in case an extractor needs to use internal state it should not affect other recipe runs. (esp when we are running multiple recipes asynchronously)

StewartJingga: it allows me to do this thing [here](https://github.com/odpf/meteor/pull/149/files?file-filters%5B%5D=.go#diff-6157bb7014e75b61c5b682ac904d45bf3d64be94cf695cc1a8c4d40278b7ab0bR30).

I hope by allowing extractor to utilise its internal states, it could lead to better code, esp in readability part.
And by making sure factory always returns a new instance, I dont have to worry about conflicting states.

wdyt?

kushsharma: Got it, makes sense. So having an independent internal state for each could be useful but I don't think the example that you gave here is great. We could have just passed it as an argument right?

StewartJingga: in this case yes I agree, we totally could pass it as an argument and it would be fine, but I'm hoping that could set an example to utilise internal states.

Also on the edge case that the extractor so complex that requires a lot of function callings, we wouldn't have to keep passing the channel everytime.

ravisuhag: +1 to extractor internal states.

docker-compose.yml
```diff
@@ -18,10 +18,3 @@ services:
       - ALLOW_PLAINTEXT_LISTENER=yes
```
ravisuhag, line 18: @StewartJingga Do we need docker-compose anymore for this?

StewartJingga: for kafka we still need it because we need two containers running, zookeeper and kafka.
And they need to be in the same network (which is being taken care of in docker-compose).

right now dockertest is helpful for spinning up a single container, but since testing kafka requires two containers, I havn't successfully spinned up two containers and connect them in a single network hence I still need docker-compose for this case.

I tried this approach [here](https://github.com/ory/dockertest/issues/195#issuecomment-775895582) mentioned in the dockertest issues but still unable to connect.

ravisuhag: We can use the new version of Kafka which is without zookeeper.

StewartJingga: we could but there are two concerns here:
1. The latest official kafka docker image is 2.8.0 which is also the same version where KRaft (Kafka Raft mode to replace zookeeper) is introduced. They put a warning [here](https://github.com/apache/kafka/blob/2.8/config/kraft/README.md#warning) that it is still for testing and would be production ready on `v3.0`.
2. We would need to create our own docker image to enable KRaft mode since a few configurations have not been exposed by the latest docker image and need to be tweaked manually.

A few users have created a kafka image with KRaft mode enabled which we can directly use, but I'm not sure it is a good idea to use that image since mostly it is for their use cases and we can't guarantee stability.

ravisuhag: Since it is for integration tests. I think it should be fine. Later we can move it to 3.0.

StewartJingga: ok sure, then we need create our own docker image to enable KRaft mode.

Do we have a repo to store our Dockerfiles?

ravisuhag: is there any public build image with Kraft that we can use?

StewartJingga: Maybe we could put the dockerfile in the meteor repo for now since atm it is only being used by its kafka extractor.
e.g.
```
plugins/extractors/kafka/Dockerfile
```

wdyt?

StewartJingga: > is there any public build image with Kraft that we can use? if yes, let's use that. If not then it is ok for us to wait till it is released.

found one [here](https://www.morling.dev/blog/exploring-zookeeper-less-kafka/).

StewartJingga: oh wait, even that one is not yet published to dockerhub, we still need to build the image ourselves 😞 .

Found another one here in dockerhub https://hub.docker.com/r/moeenz/docker-kafka-kraft.
It has `141 Downloads` and `0 stars` as of now. It probably works fine, I can try it, but are you positive with using unverified public image?

ravisuhag: Since this is only for integration testing. it should be fine. We will move to a more stable image when released.

PR #150, opened 2021-07-27
refactor: kafka sink to serialize message as protobuf

kushsharma:
Converting kafka sink to serialize messages to proto by default instead of json
- for now this sink doesn't push kafka key, only value as the proto message

### DIFF
--- a/plugins/sinks/console/sink.go
+++ b/plugins/sinks/console/sink.go
@@ -25,13 +25,11 @@ func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-ch
 }
 
 func (s *Sink) Process(value interface{}) error {
-	data, err := json.Marshal(value)
+	jsonBytes, err := json.Marshal(value)
 	if err != nil {
-		fmt.Println(fmt.Sprint(value))
-		return nil
+		return err
 	}
-	fmt.Println(string(data))
-
+	fmt.Println(string(jsonBytes))
 	return nil
 }
 
--- a/plugins/sinks/kafka/sink.go
+++ b/plugins/sinks/kafka/sink.go
@@ -2,79 +2,68 @@ package kafka
 
 import (
 	"context"
-	"encoding/json"
-	"errors"
-	"fmt"
-
 	"github.com/confluentinc/confluent-kafka-go/kafka"
+	"github.com/golang/protobuf/proto"
 	"github.com/odpf/meteor/core"
 	"github.com/odpf/meteor/core/sink"
+	"github.com/odpf/meteor/utils"
+	"github.com/pkg/errors"
 )
 
 type Sink struct{}
-
-func init() {
-	if err := sink.Catalog.Register("kafka", New()); err != nil {
-		panic(err)
-	}
-}
 
 func New() core.Syncer {
 	return new(Sink)
 }
 
-// Kafka sink
-func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-chan interface{}) (err error) {
+type Config struct {
+	Brokers string `mapstructure:"brokers" validate:"required"`
+	Topic string `mapstructure:"topic" validate:"required"`
+}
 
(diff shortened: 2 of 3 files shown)


plugins/sinks/console/sink.go
```diff
@@ -25,13 +25,12 @@ func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-ch
 }
 
 func (s *Sink) Process(value interface{}) error {
-	data, err := json.Marshal(value)
+	jsonBytes, err := json.Marshal(value)
 	if err != nil {
 		fmt.Println(fmt.Sprint(value))
```
StewartJingga, line 30: Btw why do we contain the error here? 🤔

kushsharma: Haha this is not my change although I should have cleaned it up, I only renamed the variable while resolving merge conflicts. Let me fix it.

plugins/sinks/kafka/sink.go
```diff
@@ -2,79 +2,67 @@ package kafka
...
-	}
-}
-
 func New() core.Syncer {
 	return new(Sink)
 }
 
-// Kafka sink
 func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-chan interface{}) (err error) {
+	if config["brokers"] == nil {
```
StewartJingga, line 19: for handling config, you can use this [utility](https://github.com/odpf/meteor/blob/main/utils/config.go) to help you build, validate and use default values.

sample usage can be found here:
https://github.com/odpf/meteor/blob/main/plugins/extractors/mysql/extractor.go#L38

as for now the error message is not being output clearly which Im currently working on right now.

kushsharma: Got it, using the util!

plugins/sinks/kafka/sink.go
```diff
@@ -2,79 +2,68 @@ package kafka
...
-}
-
 func New() core.Syncer {
 	return new(Sink)
 }
 
-// Kafka sink
-func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-chan interface{}) (err error) {
+type Config struct {
+	Brokers string `mapstructure:"brokers"`
```
StewartJingga, line 20: super sorry for not being clear, you need to define if the field if required using tag like
```
Brokers string `mapstructure:"brokers" validate: "required"`
```

Sample:
https://github.com/odpf/meteor/blob/main/plugins/extractors/mysql/extractor.go#L23

plugins/sinks/kafka/sink.go
```diff
@@ -2,79 +2,68 @@ package kafka
...
+	Topic string `mapstructure:"topic"`
+}
 
-	producer, err := getProducer(config)
-	if err != nil {
+func (s *Sink) Sink(ctx context.Context, config map[string]interface{}, out <-chan interface{}) (err error) {
```
StewartJingga, line 24: would `in <-chan interface{}` be better here?

plugins/sinks/kafka/sink.go
```diff
@@ -2,79 +2,68 @@ package kafka
...
+	producerConf := &kafka.ConfigMap{}
+	producerConf.SetKey("bootstrap.servers", brokers)
+	producerConf.SetKey("acks", "all")
+	return kafka.NewProducer(producerConf)
 }
 
-func (s *Sink) Push(producer *kafka.Producer, deliveryChan chan kafka.Event, value interface{}, config map[string]interface{}) error {
-	data, err := json.Marshal(value)
-
+func (s *Sink) Push(producer *kafka.Producer, value interface{}, topic string) error {
```
StewartJingga, line 49: should we use lowercase `push` as the method name just to quickly show that the function is private and not being used elsewhere.

StewartJingga (changes requested):
add required tag on config struct to make sure config is valid.

StewartJingga (approved):

PR #151, opened 2021-07-28
feat: kafka sink should populate key from reflection

kushsharma:
- protobuf used to create the value will be used to generate the kafka key

```
name: date-kafka-recipe
source:
  type: date
sinks:
  - name: kafka
    config:
      brokers: localhost:9092

      # kafka topic name
      topic: test-3

      # Use this path to find value that will be used to set kafka key proto
      # message.
      # Only supports root level key at the moment
      # Should be a string
      key_path: ".Action"
```

### DIFF
--- a/examples/date-kafka.yaml
+++ b/examples/date-kafka.yaml
@@ -5,4 +5,12 @@ sinks:
   - name: kafka
     config:
       brokers: localhost:9092
-      topic: test-1
+
+      # kafka topic name
+      topic: test
+
+      # Use this path to find value that will be used to set kafka key proto
+      # message.
+      # Only supports root level key at the moment
+      # Should be a string
+      key_path: ".Action"
--- a/go.mod
+++ b/go.mod
@@ -48,4 +48,5 @@ require (
 	google.golang.org/api v0.47.0
 	google.golang.org/protobuf v1.26.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
+	gotest.tools/v3 v3.0.2
 )
--- a/plugins/extractors/date/date.go
+++ b/plugins/extractors/date/date.go
@@ -2,12 +2,13 @@ package date
 
 import (
 	"context"
-	"fmt"
 	"time"
+
+	"github.com/odpf/meteor/proto/odpf/meta/common"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/odpf/meteor/core/extractor"
 	"github.com/odpf/meteor/plugins"
-	"github.com/odpf/meteor/proto/odpf/meta/facets"
 )
 
 const (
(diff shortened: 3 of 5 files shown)


kushsharma:
All of this is without stencil. We can add stencil support as well to avoid providing `.proto` files maybe in stencil JAM. 🕺

StewartJingga:
neat!!

Im thinking if we should allow user to use a default proto key defined by us.
since right now Meteor handles it owns proto model, technically we could also decide the proto key from that model.

and if user wants to override it, we let them use those configs.

this would also allow us to not having to define the proto key everytime on every recipe that sinking to kafka.
wdyt? @kushsharma

ravisuhag:
> neat!!
> 
> Im thinking if we should allow user to use a default proto key defined by us.
> since right now Meteor handles it owns proto model, technically we could also decide the proto key from that model.
> 
> and if user wants to override it, we let them use those configs.
> 
> this would also allow us to not have to define the proto key every time on every recipe that sinking to Kafka.
> wdyt? @kushsharma

+1 Do we need even need the proto file and name provided. Meteor already has the struct. It does not have to worry about unmarshalling proto. unless some extractor is marshaling the data and sending bytes.

kushsharma:
> neat!!
> 
> Im thinking if we should allow user to use a default proto key defined by us.
> since right now Meteor handles it owns proto model, technically we could also decide the proto key from that model.
> 
> and if user wants to override it, we let them use those configs.
> 
> this would also allow us to not having to define the proto key everytime on every recipe that sinking to kafka.
> wdyt? @kushsharma

Do you mean having single protobuf for all kafka keys in messages?

kushsharma:
> > neat!!
> > Im thinking if we should allow user to use a default proto key defined by us.
> > since right now Meteor handles it owns proto model, technically we could also decide the proto key from that model.
> > and if user wants to override it, we let them use those configs.
> > this would also allow us to not have to define the proto key every time on every recipe that sinking to Kafka.
> > wdyt? @kushsharma
> 
> +1 Do we need even need the proto file and name provided. Meteor already has the struct. We can just provide the cofigs for folks to override the keys. It does not have to worry about unmarshalling proto. unless some extractor is marshaling the data and sending bytes.

What meteor struct are you referring to? Sorry didn't get it, can you give an example?

StewartJingga:
> > neat!!
> > Im thinking if we should allow user to use a default proto key defined by us.
> > since right now Meteor handles it owns proto model, technically we could also decide the proto key from that model.
> > and if user wants to override it, we let them use those configs.
> > this would also allow us to not having to define the proto key everytime on every recipe that sinking to kafka.
> > wdyt? @kushsharma
> 
> Do you mean having single protobuf for all kafka keys in messages?

right now in our metadata model, we also have proto key defined [here](https://github.com/odpf/meteor/blob/main/proto/odpf/meta/Table.pb.go#L25).

we could use that, or @rohilsurana actually brought up an idea that we could use the model message proto as the proto key itself.

So when building the key, we would use the same proto that we use for the metadata, but instead of populating all the value, we would only populate fields that we want to mark as the key (default to Urn).

kushsharma:
> So when building the key, we would use the same proto that we use for the metadata, but instead of populating all the value, we would only populate fields that we want to mark as the key (default to Urn).

I like this idea, let me do this instead as the default behavior of autogenerating a new message from scratch. I guess we can keep the `.proto` option for users who want to use a different proto for key?

StewartJingga:
> I guess we can keep the `.proto` option for users who want to use a different proto for key?

Since it is already there I guess we can keep it, it would be helpful if we decided to introduce custom proto for sinking to kafka.

kushsharma:
Yeah, i am also okay with having only one way. We can choose to remove this custom proto option until someone requests it

ravisuhag:
@kushsharma  @StewartJingga LGTM

ravisuhag (approved):
LGTM

StewartJingga (approved):

PR #153, opened 2021-08-03
refactor: cleanup and organse cli commands

ravisuhag:
- Organise cli commands 
- Unified interface to run recipe for file and directory.


cmd/run/run.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Run a single recipe
+func run(recipeFile string) {
+	c, err := config.LoadConfig()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log := logger.New(c.LogLevel)
+	plugins.Log.Level = log.Level
```
StewartJingga, line 48: should we hide this Log level update behind a function?
e.g.
```
plugins.SetLogLevel(log.Level)
```

cmd/run/run.go
```diff
@@ -0,0 +1,83 @@
+package run
```
StewartJingga, line 1: having a package for a command should be fine, but I was just thinking if it is an overkill 🤔 

what is the reason for a command to have its own package?

cmd/run/run.go
StewartJingga: or maybe we can even pass the logger to the plugins so we would be using the same instance.
e.g.
```
plugins.SetLog(log)
```

ravisuhag: Yes, we can do that. Also, we are using native logger and logrus both. Didn't change it this PR but we should use just one of them.

cmd/run/run.go
ravisuhag: It allows easy navigation of commands. When the commands within commands will grow. As an ex. gen will multiple of these.

cmd/run/run.go
StewartJingga: oh thought it was changed this PR, tried my proposed changes in my local and had some issues due to how we register our plugins now.

let's keep it this way first then.

cmd/run.go
```diff
@@ -22,8 +56,56 @@ func run(recipeFile string) {
 	if err != nil {
 		log.Fatal(err)
 	}
+	fmt.Println(rcp)
```
StewartJingga, line 59: is this intended?

ravisuhag: No, removed.

StewartJingga (approved):

PR #154, opened 2021-08-03
feat: add metabase extractor

GrayFlash:
## Deliverables

- [x] add unit tests
- [x] add extractor
- [x] add README.md in ```plugins/extractors/metabase```, defining output
- [x] register your extractor ```plugins/extractors/populate.go```
- [x] add extractor the extractor list in ```docs/reference/extractor.md```

## Outputs

| Field | Sample Value |
| :---- | :---- |
| `urn` | `metabase.dashboard_name` |
| `name` | `dashboard_name` |
| `source` | `metabase` |
| `description` | `table description` |
| `schema` | [][Chart](#chart) |

### Chart (Basically the cards in metabase)

| Field | Sample Value |
| :---- | :---- |
| `urn` | `metabase.dashboard_name.card_name` |
| `source` | `metabase` |
| `dashboard_urn` | `metabase.dashboard_name` |
| `description` | `card description` |


plugins/extractors/metabase/extractor.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/odpf/meteor/plugins"
+	meteorMeta "github.com/odpf/meteor/proto/odpf/meta"
+	"github.com/odpf/meteor/utils"
+)
+
+var (
+	client = &http.Client{
+		Timeout: 2 * time.Second,
+	}
+	session_id = ""
```
StewartJingga, line 26: what is this value for? btw in the real application, this extractor will be instantiated multiple times, will this package variable be an issue?

plugins/extractors/metabase/extractor.go
```diff
@@ -0,0 +1,213 @@
...
+		dashboardInfo, err := e.getDashboardInfo(strconv.Itoa(dashboard.ID), cfg, dashboard.Name)
+		if err != nil {
+			return err
+		}
+		Charts := make([]*meteorMeta.Chart, len(dashboardInfo.Charts))
+		for i, chart := range dashboardInfo.Charts {
+			c := e.metabaseCardToMeteorChart(chart.Source, chart.DashboardUrn, chart.Urn)
+			Charts[i] = &c
+		}
+		e.out <- meteorMeta.Dashboard{
```
StewartJingga, line 101: since we use the `out` channel here, do we still need to store it as a private field?

plugins/extractors/metabase/extractor.go
```diff
@@ -0,0 +1,213 @@
...
+func (e *Extractor) metabaseCardToMeteorChart(source, dashboardUrn, Urn string) meteorMeta.Chart {
+	return meteorMeta.Chart{
+		Urn:             Urn,
+		Source:          source,
+		DashboardUrn:    dashboardUrn,
+		DashboardSource: "metabase",
+	}
+}
+
+func (e *Extractor) getDashboardInfo(id string, cfg Config, name string) (data Dashboard, err error) {
```
StewartJingga, line 121: let's just directly map the response to `meta.Dashboard` in this function.

plugins/extractors/metabase/extractor_test.go
```diff
@@ -0,0 +1,341 @@
...
+		req.Header.Set("X-Metabase-Session", session_id)
+	}
+	res, err = client.Do(req)
+	if err != nil {
+		return
+	}
+	return
+}
+
+func unmarshalResponse(res *http.Response, data interface{}) (err error) {
```
StewartJingga, line 334: we can merge this function directly to `newRequest` (also we should rename it to `sendRequest` since we are immediately do the request).
e.g.
```
func sendRequest(method, url string, data interface{}) (err error) {}
```

StewartJingga: same applied to `unmarshalResponse` method in extractor

GrayFlash: Got it, but it will require 4 parameters right??

```func sendRequest(method, url string, body io.Reader, data interface{}) (err error) {}```

File: plugins/extractors/metabase/extractor.go (hunk 1 above)
Comment (GrayFlash) [reply plugins/extractors/metabase/extractor.go:26]: Before making the first request to a particular api, in Metabase we need to obtain a session id with our credentials. So I used it as a global variable to avoid passing it with every function.

I am not sure if it should be a problem

Comment (StewartJingga) [reply plugins/extractors/metabase/extractor.go:26]: there is a possibility that this extractor will be instantiated multiple times and will share this `session_id` variables.
You could utilise extractor internal state (private fields) to avoid passing it to every function.
e.g.
```
type Extractor {
  sessionID string
}
```

File: plugins/extractors/metabase/extractor_test.go (hunk 4 above)
Comment (StewartJingga) [reply plugins/extractors/metabase/extractor_test.go:334]: we wouldnt need `body io.Reader` anymore.
basically the method will create a request for you and map the result to `data interface{}`.

`body io.Reader` is just the implementation of the method.

File: plugins/extractors/metabase/extractor.go [hunk 5]
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Extract collects metdata from the source. Metadata is collected through the out channel
+func (e *Extractor) Extract(ctx context.Context, config map[string]interface{}, out chan<- interface{}) (err error) {
+	// Build and validate config received from receipe
+	if err := utils.BuildConfig(config, &e.cfg); err != nil {
+		return extractor.InvalidConfigError{}
+	}
+	// get session id for further api calls in metabase
+	if e.cfg.SessionID == "" {
```
StewartJingga, line 53: since we already have `getSessionID` method, we could move this `if` logic to that function instead.
it should clean a few lines in this function.

GrayFlash: So we should call function ```getSessionID``` directly and make the session id's request as per this condition??

StewartJingga: yeah basically `getSessionID` will be responsible for creating sessionID. This function will just call it and expect a sessionID without caring how to generate it.

StewartJingga: i would something like
```
e.sessionID = e.getSessionID()
```

plugins/extractors/metabase/extractor.go
```diff
@@ -0,0 +1,160 @@
...
+	err = e.makeRequest("POST", e.cfg.Host+"/api/session", values, data)
+	if err != nil {
+		return
+	}
+	e.sessionID = data.ID
+	return
+}
+
+// helper function to avoid rewriting a request
+func (e *Extractor) makeRequest(method, url string, values map[string]interface{}, data interface{}) (err error) {
```
StewartJingga, line 125: we could rename `values` to `payload` and it should be an interface instead of a map since we could also pass a struct if needed.

StewartJingga (approved):

plugins/extractors/metabase/extractor.go
```diff
@@ -0,0 +1,160 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"time"
+
+	_ "github.com/lib/pq"
+	"github.com/odpf/meteor/core"
```
ravisuhag, line 14: @GrayFlash This package has been moved. Check main branch and you would need to make some changes in this.

PR #155, opened 2021-08-04
fix: stuck process on error

StewartJingga:
- Modify streaming logic to prevent stuck process on error
- Register plugins to be discovered by recipe
- Fix `enrich` processor to handle streaming data


cmd/run.go
```diff
@@ -78,11 +79,11 @@ func rundir(dirPath string) {
 	log := logger.New(c.LogLevel)
 	plugins.Log.Level = log.Level
 
-	failedRecipes, err := initRunner(c, log).RunMultiple(recipeList)
+	runs := initRunner(c, log).RunMultiple(recipeList)
 	if err != nil {
```
ravisuhag, line 83: with err removed is this check needed?

core/recipe/runner.go
```diff
@@ -30,192 +30,150 @@ func NewRunner(
...
+		}()
+
+		// replace the channel with the new out channel
+		// this would allow the next processor or sink to
+		// receive the processed data instead of data directly from extractor
+		channel = outChannel
+	}
+
+	// run sinks
+	for _, sinkRecipe := range recipe.Sinks {
```
ravisuhag, line 88: Should we run sinks first. Because if sinks are not ready there is no point of running the extractors. Order should be sinks > processors > extractors.

core/recipe/runner.go
```diff
@@ -30,192 +30,150 @@ func NewRunner(
...
+	var (
+		getDuration = r.startDuration()
+		channel     = make(chan interface{})
+	)
+	run.Recipe = recipe
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
+	// run extractors
+	{
```
ravisuhag, line 51: Any reason for this scope?

StewartJingga: no big reason, purely for readability. If it does not help I can remove it.

ravisuhag: Yes, I think it just makes the reader think and look for reasons and creates confusion.

core/recipe/runner.go
StewartJingga: we can't due to channels dependencies in this current logic.

say we have a recipe with 1 extractor (`extr-a`), 2 processors (`proc-a` & `proc-b`), and 2 sinks (`sink-a`, `sink-b`).
then we would have below structure:

```
extr-a --> channel-1 --> proc-a
proc-a --> channel-2 --> proc-b
proc-b --> channel-3 --> [sink-a, sink-b]
```

so basically I need to define the inputter of the channel before the listener of the channel.

technically the order of the execution won't matter much since they are all running asynchronously, so it is totally possible to setup sinks first.
but it would make the channels management to be much much harder.

> Because if sinks are not ready there is no point of running the extractors

yes, but maybe we can also look it this way, `if there is no extractor, there is nothing to be done and to listen to`.
Even processors would be useless without extractors. Also `extractor -> processors -> sinks` is actually what is going on. wdyt?

ravisuhag: I think it is fine for now. Later we can think of having an init function in each extractor and processor that can run first before starting extraction to check all connections etc.

PR #157, opened 2021-08-09
refactor: extract agent and recipe packages

ravisuhag:
- Extract agent as a package that will take care of running recipe. 
- Extract recipe as a core package 
- Move metrics and logger to internal


StewartJingga (commented):
Fully onboard with separating runner with recipe package.

Im just still deciding whether `agent` is a good representation. 🤔

ravisuhag:
> Fully onboard with separating runner with recipe package.
> 
> Im just still deciding whether `agent` is a good representation. 🤔

Agent is someone who acts on behalf. Here agent is acting on behalf of the meteor to run the whole program and it takes care of moving data around as well. That's the reason to name it agent. Any other suggestions?

PR #160, opened 2021-08-11
feat(columbus): add columbus sink

StewartJingga:
- add a new sink - `columbus`

### Sink Config
| Field | Sample Value |
| :---- | :---- |
| `host` | `https://columbus.com` |
| `type` | `sample-columbus-type` |


plugins/sinks/columbus/sink.go
```diff
@@ -0,0 +1,92 @@
...
+			return
+		}
+
+		err = fmt.Errorf("columbus returns %d: %v", res.StatusCode, string(bodyBytes))
+	}
+
+	return
+}
+
+func (s *Sink) buildPayload(data interface{}) (payload []byte, err error) {
```
GrayFlash, line 80: Is this function really needed, cause it is being used once? And it is not to be to spoil readability.

StewartJingga: Oh actually it is not for reusability. I wanted to hide this from `sink()` method.
```
columbusPayload := []interface{}{data}
```
It may look confusing without context and I had to put comments to make it clear.
Basically I want them just to read it as `buildPayload`.

GrayFlash: Oh okay

GrayFlash: Other than this the code looks good to me

plugins/sinks/columbus/sink.go
```diff
@@ -0,0 +1,92 @@
...
+	for data := range in {
+		if err = s.sink(data); err != nil {
+			return
+		}
+	}
+
+	return
+}
+
+func (s *Sink) sink(data interface{}) (err error) {
```
ravisuhag, line 49: Let's rename it from sink to something else maybe send it to avoid confusion with Sink method.

ravisuhag:
@StewartJingga Recent merge will require some changes on this.

PR #162, opened 2021-08-12
refactor: registry package

ravisuhag:
- Extract registry package 
- Remove core package


registry/errors.go
```diff
@@ -1,11 +1,12 @@
-package processor
+package registry
 
 import "fmt"
 
 type NotFoundError struct {
```
StewartJingga, line 5: since `registry` package is basically just plugins factories and the core definitions of plugins is in `plugins` package, should we move this error to `plugins` package instead?

StewartJingga: also following up on this, I'm thinking if we can just put these factories inside `plugins` and remove `registry` package.
wdyt?

ravisuhag: Yes, thinking the same. But going forward we might need to do a lot more with the registry. getting info of a single plugin, displaying all extractors in table format using the list command. With that in mind kept it like this. 
For error keeping it here as this error is related to the registry only. plugin does not care about it. What do you think?

StewartJingga: > Yes, thinking the same. But going forward we might need to do a lot more with the registry.

ack

> For error keeping it here as this error is related to the registry only. plugin does not care about it.

since the definitions are now in the `plugins` I think it makes sense to put error definition in the same package.
it is like defining an arbitrary core model with errors in the same package even though the one who will be using/producing the errors is the `store/repo` package.

Aside from having all of our plugins errors in one place, I feel like `plugins.NotFoundError` looks nice.
wdyt?

ravisuhag: Make sense. Made the change.

StewartJingga (approved):
LGTM

PR #164, opened 2021-08-13
feat(columbus): allow mapping fields when sinking to columbus

StewartJingga:
- Add a new feature in `columbus` sink to map model fields to custom fields when sinking to columbus.

### Config
| Field | Sample Value |
| :---- | :---- |
| `fields_mapper` | `{"ModelField":"new_fieldname"}` |


plugins/sinks/columbus/sink.go
```diff
@@ -33,27 +35,73 @@ func New(c httpClient) plugins.Syncer {
...
+	sf := make([]reflect.StructField, 0)
+	for i := 0; i < t.NumField(); i++ {
+		sf = append(sf, t.Field(i))
+
+		// check if field will be mapped
+		jsonField, ok := fieldMap[t.Field(i).Name]
+		if !ok {
+			continue // skip if field is not mentioned in the mapper
+		}
+		sf[i].Tag = reflect.StructTag(fmt.Sprintf(`json:"%s"`, jsonField))
```
ravisuhag, line 89: Can it be written concisely? maybe like this.
```
if jsonField, ok := fieldMap[t.Field(i).Name]; ok{
  sf[i].Tag = reflect.StructTag(fmt.Sprintf(`json:"%s"`, jsonField))
}

File: docs/reference/sinks.md [hunk 2]
```diff
@@ -27,4 +27,11 @@ sinks:
    config:
      host: https://columbus.com
      type: sample-columbus-type
-```
\ No newline at end of file
+     fields_mapper:
```
ravisuhag, line 30: This can be called just mapping. As it is evident that it is for fields.

plugins/sinks/columbus/sink.go
StewartJingga: oh was in a habit of exiting early, but since the valid condition is only a one liner we can do it your way.

plugins/sinks/columbus/sink.go
```diff
@@ -33,27 +35,71 @@ func New(c httpClient) plugins.Syncer {
...
 func (s *Sink) Sink(ctx context.Context, configMap map[string]interface{}, in <-chan interface{}) (err error) {
-	if err = utils.BuildConfig(configMap, &s.config); err != nil {
+	var config Config
+	if err = utils.BuildConfig(configMap, &config); err != nil {
 		return plugins.InvalidConfigError{Type: plugins.PluginTypeSink}
 	}
 
 	for data := range in {
-		if err = s.send(data); err != nil {
+		dataMapper := s.getDataMapper(data, config)
```
GrayFlash, line 44: @StewartJingga can we move config to Sink struct, just like we did in metabase extractor?? Since it is being passed as a parameter.

ravisuhag: @StewartJingga Any thoughts on the above comment? Apart from this LGTM.

StewartJingga: > @StewartJingga can we move config to Sink struct, just like we did in metabase extractor?? Since it is being passed as a parameter.

oh actually being passed a parameter is not the reason for utilising internal states, unless you find yourself passing it over and over again. But yeah sure both approach should be fine here, will update the PR.

StewartJingga: updated 🙏 @GrayFlash

GrayFlash (approved):

PR #165, opened 2021-08-14
docs: improve documentations

GrayFlash:
Deliverables:

- [x] Add detailed concepts about recipes, making them easier to be created.
- [x] Document metadata models
- [x] Feature matrix for extractors
- [x] look through references and find improvements.


StewartJingga:
should rename `docs/data models` to `docs/metadata_models`?

`models` alone without metadata can also mean `meteor models` such as `recipe`, `processors`, etc.
wdyt?

StewartJingga:
@ravisuhag having this feature matrix will require contributors to always make sure it is updated. Also whenever there are model changes we need to keep track of it as well here.

It's nice to have it but was just wondering if the maintenance outweighs the benefits. 🤔

ravisuhag:
@StewartJingga We will maintain it by versions. we can specify the version in docs. But having this will allow contributors and users to easily understand what is present.

ravisuhag:
@StewartJingga @GrayFlash To address update concerns, I have added comments in proto file of models. which can be used to generate docs for those models and we can link to them. For the feature matrix maybe we can have only a high-level facet name instead of each attribute to give clarity and more info can be linked to the readme of each extractor. Thoughts?

GrayFlash:
@ravisuhag It seems fine for me we can do that and I have actually not added all the attributes and just written facet names actually in the feature matrix. For e.g. used `Timestamp` and not `last_updated` or `last_edited` attribute.

Btw any issue around the metadata model's readme or other updates I made.

StewartJingga:
> For the feature matrix maybe we can have only a high-level facet name instead of each attribute to give clarity and more > info can be linked to the readme of each extractor.

Yeah showing only facets should be pretty descriptive of the model and would be easier to maintain.

PR #166, opened 2021-08-16
refactor: logging and recipe package

ravisuhag:
- cleanup recipe package and logging
- feat: replace generator to recipe
- refactor: use salt logger and remove the internal logger
- fix: handle status and error for multiple recipes


cmd/run.go
```diff
@@ -2,106 +2,49 @@ package cmd
...
-		return nil
-	}
+			runs := runner.RunMultiple(recipes)
 
-	client, err := metrics.NewStatsdClient(c.StatsdHost)
-	if err != nil {
-		panic(err)
+			for _, run := range runs {
+				if run.Error != nil {
+					lg.Error(fmt.Sprintf("%s: %s", run.Recipe.Name, run.Error))
```
kushsharma, line 42: ```lg.Error(run.Error.Error(), "name", run.Recipe.Name)```

cmd/run.go
```diff
@@ -2,106 +2,49 @@ package cmd
...
+					continue
+				}
+				lg.Info(fmt.Sprintf("%s: success", run.Recipe.Name))
```
kushsharma, line 45:
```
lg.Info("success", "name", run.Recipe.Name)
```
the first argument is a string explaining what the log is, rest of the args are optional and alternates at key/value pairs.
Args key should be string and value could be anything printable.

plugins/extractors/bigquery/extractor.go
```diff
@@ -60,7 +59,7 @@ func (e *Extractor) Extract(ctx context.Context, config map[string]interface{},
 			break
 		}
 		if err != nil {
-			e.logger.Error(err)
+			e.logger.Error("failed to fetch, skipping dataset", err)
```
kushsharma, line 62: this will not work as expected as the method takes the message and key/value pairs. What we should be doing is
```
e.logger.Error("failed to fetch, skipping dataset", "error", err)
```
or
```
e.logger.Error(errors.Wrap(err, "failed to fetch, skipping dataset").Error())
```
or
```
e.logger.Error(err.Error())
```

plugins/extractors/csv/extractor_test.go
```diff
@@ -5,23 +5,22 @@ package csv_test
...
 	"github.com/odpf/meteor/plugins"
 	"github.com/odpf/meteor/plugins/extractors/csv"
 	"github.com/odpf/meteor/proto/odpf/meta"
 	"github.com/odpf/meteor/proto/odpf/meta/facets"
+	logger "github.com/odpf/salt/log"
 	"github.com/stretchr/testify/assert"
 )
 
-var log = logger.NewWithWriter("info", ioutil.Discard)
+var lg = logger.NewLogrus()
```
kushsharma, line 18: do we need logger in test files? We can use `Noop` logger here

plugins/extractors/kafka/extractor_test.go
```diff
@@ -179,7 +179,7 @@ func cleanUp(ctx context.Context, client *kafkaLib.AdminClient) (err error) {
 
 func newExtractor() *kafka.Extractor {
 	return kafka.New(
-		logger.NewWithWriter("info", ioutil.Discard),
+		logger.NewLogrus(logger.LogrusWithWriter(ioutil.Discard)),
```
kushsharma, line 182: try using `noop`

plugins/extractors/postgres/postgres.go
```diff
@@ -57,19 +57,19 @@ func (e *Extractor) Extract(ctx context.Context, config map[string]interface{},
 		// information will be returned
 		db, err := connection(cfg, database)
 		if err != nil {
-			e.logger.Error(err)
+			e.logger.Error("failed to connect,skipping database", err)
```
kushsharma, line 60: `connect,skipping`

plugins/logger.go
```diff
@@ -1,14 +1,14 @@
 package plugins
 
-import "github.com/odpf/meteor/internal/logger"
+import "github.com/odpf/salt/log"
 
 type Logger interface {
```
kushsharma, line 5: We can reuse salt logger [interface](https://github.com/odpf/salt/blob/main/log/logger.go#L21) right? I mean I agree we should have our own interface but because salt is managed in odpf, we know what we are supporting so I guess we can reuse?

recipe/reader.go
```diff
@@ -4,28 +4,45 @@ import (
...
-	reader.populateData()
+	reader.data = populateData()
 
 	return reader
 }
 
-func (r *Reader) Read(path string) (recipe Recipe, err error) {
+//  Read loads the list of recipes from a give file or directory path.
+func (r *Reader) Read(path string) (recipes []Recipe, err error) {
+	fi, err := os.Stat(path)
```
kushsharma, line 25: Small suggestion, although current implementation is great we can also think of using a file system interface abstraction like [afero](https://github.com/spf13/afero). This will also allow in-memory filesystem really helpful in testing.

recipe/envs.go
```diff
@@ -0,0 +1,44 @@
...
+var (
+	recipeEnvVarPrefix = "METEOR_"
+)
+
+func populateData() map[string]string {
+	data := make(map[string]string)
+	for _, envvar := range os.Environ() {
+		keyval := strings.SplitN(envvar, "=", 2) // "sampleKey=sample=Value" returns ["sampleKey", "sample=value"]
+		key := keyval[0]
+		val := keyval[1]
```
kushsharma, line 17: Should we expand `value` using `os.ExpandEnv` as well?
If someone provides an env like `output=$HOME`, I think value needs to be expanded to work although I am not sure if that is one of the possible use cases users can use.

cmd/run.go
```diff
@@ -2,106 +2,49 @@ package cmd
...
+			reader := recipe.NewReader()
+			recipes, err := reader.Read(args[0])
+			if err != nil {
+				return err
+			}
 
-	runs := initRunner(c, log).RunMultiple(recipeList)
-	log.WithField("runs", runs).Info("Done!")
-}
+			count := len(recipes)
```
kushsharma, line 31: few rules I personally follow to decide if we should create a new variable in these cases
1. Is this value getting used in more than one place? If no, don't create
2. Is creating a variable out of it helps create an alias and them improves code readability? If no, and its already very obvious to read, don't create
3. Is creating a new variable cache a compute-intensive operation result? If no, don't create

So for example:
```
reader := recipe.NewReader()
recipes, err := reader.Read(args[0])
if err != nil {
    return err
}
```
can be written as
```
recipes, err := recipe.NewReader().Read(args[0])
if err != nil {
    return err
}
```
and 
```
if len(recipes) == 0 {
    //..
}
```

plugins/logger.go
ravisuhag: Yes, Plan to remove this in the next PR.

cmd/run.go
ravisuhag: Agree. I was planning to use count for printing more things like how many recipes ran etc. Moved that scope to next PR.

cmd/run.go
```diff
@@ -2,106 +2,49 @@ package cmd
...
-	runner = agent.NewAgent(
-		registry.Extractors,
-		registry.Processors,
-		registry.Sinks,
-		metricsMonitor,
-	)
-	return
-}
+			if count == 0 {
+				lg.Info(fmt.Sprintf("no recipe found for path: %s", args[0]))
```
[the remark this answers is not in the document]
ravisuhag: Made the change

plugins/extractors/bigquery/extractor.go
ravisuhag: made the change

PR #170, opened 2021-08-23
refactor: QA all extractors

GrayFlash:
## QA all extractors, fix inconsistency, redundant code, unfixed registry, bug if any.

- [x] BigQuery
- [x] BigTable
- [x] Clickhouse
- [x] CSV
- [x] ElasticSearch
- [x] GCS
- [x] Github
- [x] Grafana
- [x] kafka
- [x] metabase
- [x] mongodb
- [x] mssql
- [x] mysql
- [x] postgres


GrayFlash:
@StewartJingga @ravisuhag while refactoring Metabase Extractor I renamed `host` in config to `url` because we are using metabase by making api calls on the url which is different from setting tcp connection on a given host. Seems fine right, made changes in adjacent README of metabase as well??

plugins/extractors/clickhouse/clickhouse.go
```diff
@@ -37,12 +46,15 @@ func (e *Extractor) Extract(ctx context.Context, configMap map[string]interface{
 	if err != nil {
 		return
 	}
-	result, _ := e.getTables()
-	out <- result
+	err = e.getTables()
```
StewartJingga, line 49: let's rename this function to `extractTables`.
`getTables` implies it will return a list of tables. wdyt?

plugins/extractors/elastic/elastic.go
```diff
@@ -15,28 +18,35 @@ import (
 type Config struct {
 	User     string `mapstructure:"user"`
 	Password string `mapstructure:"password"`
-	Host     string `mapstructure:"host" validate:"required"`
+	Url      string `mapstructure:"url" validate:"required"`
```
StewartJingga, line 21: let's keep Host here since Url can also include the path.

plugins/extractors/elastic/elastic.go
```diff
@@ -46,15 +56,14 @@ func (e *Extractor) Extract(ctx context.Context, configMap map[string]interface{
 		return
 	}
 
-	result, err := e.listIndexes(client)
+	err = e.listIndexes(client)
```
StewartJingga, line 59: let's rename this to `extractIndexes`.

plugins/extractors/elastic/elastic_test.go
```diff
@@ -99,32 +102,39 @@ func TestMain(m *testing.M) {
...
-			expected := getExpectedVal()
-			assert.Equal(t, expected, extractOut)
+			close(extractOut)
+
+			assert.Nil(t, err)
 		}()
+		var results []resources.Table
+		for d := range extractOut {
+			table, ok := d.(resources.Table)
+			fmt.Println(table)
```
StewartJingga, line 128: don't forget to remove this after debugging

plugins/extractors/elastic/elastic_test.go
```diff
@@ -99,32 +102,39 @@ func TestMain(m *testing.M) {
...
+			if !ok {
+				t.Fatal(errors.New("invalid table format"))
+			}
+
+			results = append(results, table)
+		}
+		fmt.Println(results)
```
StewartJingga, line 135: this too

plugins/extractors/metabase/README.md
```diff
@@ -15,7 +15,7 @@ source:
 
 | Key | Value | Example | Description |    |
 | :-- | :---- | :------ | :---------- | :- |
-| `host` | `string` | `http://localhost:3000` | The url at which metabase is running | *required* |
+| `url` | `string` | `http://localhost:4002` | The url at which metabase is running | *required* |
```
StewartJingga, line 18: don't forget to revert this as well

plugins/extractors/metabase/metabase_test.go
```diff
@@ -35,18 +34,18 @@ const (
...
 	pass                   = "meteor_pass_1234"
-	port                   = "3000"
-	url                    = "http://localhost:3000"
+	port                   = "4002"
 )
 
 var (
 	client = &http.Client{
-		Timeout: 2 * time.Second,
+		Timeout: 4 * time.Second,
```
StewartJingga, line 42: was there an issue with 2 secs timeout?

plugins/extractors/mysql/mysql_test.go
```diff
@@ -38,16 +41,16 @@ func TestMain(m *testing.M) {
 		Env: []string{
 			"MYSQL_ALLOW_EMPTY_PASSWORD=true",
 		},
-		ExposedPorts: []string{"3306"},
+		ExposedPorts: []string{"3306", port},
 		PortBindings: map[docker.Port][]docker.PortBinding{
 			"3306": {
```
StewartJingga, line 46: should we update this map key as well?

StewartJingga (changes requested):

plugins/extractors/mysql/mysql_test.go
GrayFlash: No, key is the one that needs the default port and it maps it to the `port` we provide.

plugins/extractors/clickhouse/clickhouse.go
GrayFlash: Yes, makes more sense

plugins/extractors/metabase/metabase_test.go
GrayFlash: Not sure actually, but was getting retry deadline on my device so tried increasing the timeout. Shouldn't be a problem right??

StewartJingga: hmm 🤔 we may need to look into that, but let's just keep it for now.

plugins/extractors/metabase/README.md
```diff
@@ -6,7 +6,7 @@
 source:
   type: metabase
   config:
-    host: http://localhost:3000
+    url: http://localhost:3000
```
StewartJingga, line 9: don't forget to update the sample usage as well

StewartJingga (changes requested):

StewartJingga (approved):
LGTM

PR #171, opened 2021-08-23
refactor(log): switch to odpf/salt logger

StewartJingga:
- update plugins log to use odpf/salt logger
- fix infinite loop on bigquery extractor if access is authorized
- log recipe details on debug log level

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -31,8 +31,8 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor) *cobra.Command {
 			}
 
 			runs := runner.RunMultiple(recipes)
-
 			for _, run := range runs {
+				lg.Debug("recipe details", "config", run.Recipe)
 				if run.Error != nil {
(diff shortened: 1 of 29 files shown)


cmd/run.go
```diff
@@ -31,8 +31,8 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor) *cobra.Command {
 			}
 
 			runs := runner.RunMultiple(recipes)
-
 			for _, run := range runs {
+				lg.Debug("recipe details", "config", run.Recipe)
```
GrayFlash, line 35: @StewartJingga any reason we are using the debug now while earlier we weren't. Just wanted to know if odpf/salt has brought any change that is being addressed here.

GrayFlash:
Other than this LGTM!

cmd/run.go
StewartJingga: odpf/salt has nothing to do with it, the reason is because I need information about the recipes that are being run but only when log level is debug.

GrayFlash: oh okay, got it

PR #174, opened 2021-08-24
fix: build workflow to use goreleaser to build and push image

GrayFlash:


GrayFlash:
@StewartJingga build works now, tested in commit `e1ee0dc`.

PR #180, opened 2021-08-28
refactor: update metadata models

StewartJingga:
- change metadata models to latest (https://github.com/odpf/proton/tree/main/odpf/assets)
- fix breaking changes from metadata models
- update columbus sink to allow mapping nested fields


plugins/sinks/columbus/sink.go
```diff
@@ -41,57 +41,80 @@ func (s *Sink) Sink(ctx context.Context, configMap map[string]interface{}, in <-
...
 		}
 	}
 
 	return
 }
 
-func (s *Sink) getDataMapper(data interface{}) func(interface{}) interface{} {
-	if s.cachedDataMapper == nil {
-		s.cachedDataMapper = s.buildDataMapper(data)
+func (s *Sink) mapData(data interface{}) (interface{}, error) {
```
StewartJingga, line 56: @ravisuhag We need this function to map fields (e.g. `resource.urn` to `urn`) so Columbus will accept the payload.

I gave myself a time to find a better approach but I could not orchestrate a better way to map these **nested fields** before sending to Columbus without making it overly complicated. I tried a few efforts using `reflect` library and it made it super complicated and without success.

So now I'm using the most easier approach just to get it working.

ravisuhag: Works for now. Anyway, we are likely to change Columbus contract and there might be some changes in models as well.

PR #182, opened 2021-08-28
feat: add custom help

ravisuhag:
Add custom help message with command grouping capability. (Inspired from Github cli.)

### DIFF
--- a/cmd/gen.go
+++ b/cmd/gen.go
@@ -13,7 +13,7 @@ func GenCmd(lg log.Logger) *cobra.Command {
 		Use:   "gen",
 		Short: "A collection of several useful generators.",
 		Annotations: map[string]string{
-			"IsCore": "true",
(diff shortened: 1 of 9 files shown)


kushsharma:
This is nice, LGTM. Just one small thing, I noticed we are using two different conventions to add annotations
```
"IsCore": "true",
"help:feedback": "...",
```
if this annotation is for grouping core commands then maybe `group:core`? or maybe make it a const variable?

kushsharma (approved):
This is nice, LGTM. Just one small thing, I noticed we are using two different conventions to add annotations
```
"IsCore": "true",
"help:feedback": "...",
```
if this annotation is for grouping core commands then maybe `group:core`? or maybe make it a const variable?

ravisuhag:
@kushsharma made the change.

PR #185, opened 2021-08-28
feat: use salt for config

ravisuhag:


main.go
```diff
@@ -23,9 +24,10 @@ const (
 
 func main() {
 
-	cfg, err := config.Load()
+	var cfg agent.Config
+	l := config.NewLoader()
```
rohilsurana, line 28: After https://github.com/odpf/salt/issues/6 we will need to update this to include the default path for config file.
Thanks @burntcarrot !

main.go
```diff
@@ -5,10 +5,11 @@ import (
 	"os"
 	"strings"
 
+	"github.com/odpf/meteor/agent"
 	"github.com/odpf/meteor/cmd"
-	"github.com/odpf/meteor/config"
```
rohilsurana, line 9: Need to get rid of this package

main.go
burntcarrot: I was wondering about the defaults present in `config/config.go` in my local fork, thanks for pinging me here, as now I'm aware of the removal of `config/config.go` from `meteor` (which seems straightforward) as we're moving forward with the usage of `salt`.

Anyways, thanks for the mention! 😄

PR #189, opened 2021-08-31
feat: plugin info framework

StewartJingga:
- read plugin info from yaml file (e.g. `meta.yaml`)
- expose plugin config validation so recipe can be validated before running
- generate README.md from plugin info yaml (e.g. `meta.yaml`)
-


plugins/plugin.go
```diff
@@ -10,6 +10,13 @@ const (
 	PluginTypeSink      PluginType = "sink"
 )
 
+type PluginInfo struct {
```
ravisuhag, line 13: Lets just call it info. Since package name is already plugins.

agent/agent.go
```diff
@@ -29,6 +31,21 @@ func NewAgent(ef *registry.ExtractorFactory, pf *registry.ProcessorFactory, sf *
 	}
 }
 
+func (r *Agent) Validate(rcp recipe.Recipe) error {
```
ravisuhag, line 34: Does it make sense to make it part of recipe? Because that is what we are validating.

StewartJingga: we can't because recipe package does not depend on plugins and only agent currently has all the plugin factories.

plugins/sinks/columbus/meta.yaml
```diff
@@ -0,0 +1,12 @@
+name: "Columbus HTTP"
+version: "1.0.0"
+sample_config: |-
+  sinks:
```
ravisuhag, line 4: Sinks keyword we should avoid from sample config I think. as we would need to combine sample configs from multiple sinks when people add more than one.

plugins/sinks/columbus/meta.yaml
```diff
@@ -0,0 +1,12 @@
...
+    - name: columbus
```
ravisuhag, line 5: Let's call it title since this is a more human-readable name. Also, does it make sense to move the proper name also to meta.yaml from the register function? Thoughts? any pros and cons we see?

plugins/sinks/columbus/meta.yaml
StewartJingga: yeah, forgot to remove that, we should just put the config map.

@GrayFlash need help to remove that in the next push 🙏 .

plugins/sinks/columbus/meta.yaml
StewartJingga: just to make sure, changing `info.name` to `info.title`?

StewartJingga: > Also, does it make sense to move the proper name also to meta.yaml from the register function? Thoughts? any pros and cons we see?

hmm so we fully depends on the `meta.yaml` for the plugin name as well? we could but we would be fully depend on meta.yaml then, I was thinking meta.yaml is more like an addition rather than essential.

pros:
- meta.yaml holds all information

cons:
- right now plugin info is not being evaluated or calculated unless being required, this saves us startup time. Putting proper name to `meta.yaml` will require us to always build and evaluate info for all plugins because we need that information from `meta.yaml`. And we cannot opt out of it because right now we are registering our plugins via `init` which will be called on every startup.

ravisuhag: Ok. In that case, lets' rename info.name to info.titile to clear out the diffs.

plugins/processors/enrich/meta.yaml
```diff
@@ -0,0 +1,6 @@
+name: "enrich"
+version: "1.0.0"
+Description: ""
```
ravisuhag, line 3: typo on this. Description, first letter should be small case.

PR #192, opened 2021-09-01
fix(bigquery): fix breaking on failed preview building

StewartJingga:
- fix panicing when bigquery preview is failed to build
- move testutils package to root
- add logging to give more visibility


plugins/extractors/mysql/mysql_test.go
```diff
@@ -15,10 +15,10 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/odpf/meteor/plugins"
 	"github.com/odpf/meteor/plugins/extractors/mysql"
-	"github.com/odpf/meteor/plugins/testutils"
 	"github.com/odpf/meteor/proto/odpf/assets"
 	"github.com/odpf/meteor/proto/odpf/assets/common"
 	"github.com/odpf/meteor/proto/odpf/assets/facets"
+	"github.com/odpf/meteor/testutils"
```
ravisuhag, line 21: @StewartJingga Let's rename this to just `test` package.

GrayFlash: Any reason we are not including it in `utils` package we already have in root??

StewartJingga: @ravisuhag the `test` folder breaks `make test` command and the workaround is `make -B test`.

Should we keep the `testutils` package or maybe put `test` package into `utils`?

ravisuhag: @StewartJingga more context on why make is breaking with that?

StewartJingga: If you have `test` folder while running `make test`, it would refer to that folder instead of the command you specified in Makefile.

But I just found a fix for that, we can put `test` to always be a command like this

```
# Makefile

.PHONY: test
```

PR #198, opened 2021-09-06
docs: update installation and command usage

GrayFlash:
Deliverables:

- [x] Update Installations guide
- [x] List all the commands
- [x] Give an overview of plugins and explain list plugins commands.
- [x] Create and lint recipe
- [x] Run meteor on the recipes
- [x] Deploy meteor
- [x] add reference for all the commands
- [x] add installation from binaries


GrayFlash:
FYI @ravisuhag we have Installations.md in `docs/guides` so didn't add it again in the tour, but surely listed it as a pre-requisite.

Also, wanted to know is usage.md in `docs/guides` should have as we have added the list of commands and info on them in `docs/references/commands.md` and are adding tour.

ravisuhag:
@GrayFlash I think our tour and guide are the same. We can keep the name of the guide only and structure it the way we discussed.

PR #212, opened 2021-09-13
feat(metrics): record data count

GrayFlash:
Deliverables:

- [x] add datacount variable
- [ ] logical implementation to collect count of tables, charts, extracted.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -99,6 +99,7 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 	var (
 		getDuration = r.startDuration()
 		channel     = make(chan interface{})
+		dataCount   = 0
 	)
 	run.Recipe = recipe
 
@@ -113,6 +114,7 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 		}
 
 		close(extrChannel)
+		dataCount = len(extrChannel)
 	}()
 
 	// run processors
@@ -157,10 +159,9 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 
 	success := run.Error == nil
 	durationInMs := getDuration()
-	dataCount := 0
 	r.monitor.RecordRun(recipe, durationInMs, success, dataCount)
 	if success {
-		r.logger.Info("done running recipe", "recipe", recipe.Name, "duration_ms", durationInMs, "Data Count", dataCount)
+		r.logger.Info("done running recipe", "recipe", recipe.Name, "duration_ms", durationInMs, "data_count", dataCount)
 	} else {
 		r.logger.Error("error running recipe", "recipe", recipe.Name, "duration_ms", durationInMs, "err", run.Error)
 	}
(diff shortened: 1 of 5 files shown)


agent/agent.go
```diff
@@ -157,10 +157,10 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 
 	success := run.Error == nil
 	durationInMs := getDuration()
-	r.monitor.RecordRun(recipe, durationInMs, success)
-
+	dataCount := 0
+	r.monitor.RecordRun(recipe, durationInMs, success, dataCount)
 	if success {
-		r.logger.Info("done running recipe", "recipe", recipe.Name, "duration_ms", durationInMs)
+		r.logger.Info("done running recipe", "recipe", recipe.Name, "duration_ms", durationInMs, "Data Count", dataCount)
```
StewartJingga, line 163: let's try to have the same format, instead of `"Data Count"`, we can use `"data_count"`.

agent/agent_test.go
```diff
@@ -135,7 +135,7 @@ func TestRunnerRun(t *testing.T) {
 			t.Error(err.Error())
 		}
 		monitor := new(mockMonitor)
-		monitor.On("RecordRun", validRecipe, mock.AnythingOfType("int"), true).Once()
+		monitor.On("RecordRun", validRecipe, mock.AnythingOfType("int"), true, 0).Once()
```
StewartJingga, line 138: try to use other number than `0`, this is to make sure we do not test default values.

metrics/statsd.go
```diff
@@ -29,29 +29,30 @@ func NewStatsdMonitor(client statsdClient, prefix string) *StatsdMonitor {
...
-func (m *StatsdMonitor) createMetricName(metricName string, recipe recipe.Recipe, success bool) string {
+func (m *StatsdMonitor) createMetricName(metricName string, recipe recipe.Recipe, success bool, dataCount int) string {
 	var successText = "false"
 	if success {
 		successText = "true"
 	}
 
 	return fmt.Sprintf(
-		"%s.%s,name=%s,success=%s",
+		"%s.%s,name=%s,success=%s,dataCount=%s",
```
StewartJingga, line 50: use `dataCount=%d` instead for formatting number.

and we can skip the using `strconv.Itoa`.

metrics/statsd_test.go
```diff
@@ -48,25 +51,28 @@ func TestStatsdMonitorRecordRun(t *testing.T) {
 		defer client.AssertExpectations(t)
 
 		monitor := metrics.NewStatsdMonitor(client, statsdPrefix)
-		monitor.RecordRun(recipe, duration, false)
+		monitor.RecordRun(recipe, duration, false, 0)
```
StewartJingga, line 54: try to use other number than default value (in this case `0`) to make sure it works fine.

GrayFlash: Yes, was thinking of a way currently to get the datacount first. btw on this test, since the success is false the count will be 0 right??

StewartJingga: the count can be more than 0 even when it fails.
but that should not be the point, this module does not care about that, it just focuses on collecting metrics with whatever given.

PR #213, opened 2021-09-13
feat: simplify plugins implementation using emitter

StewartJingga:
- apply new Record model to wrap extracted data
- simplify plugins implementation using emitter


StewartJingga:
@ravisuhag changed emitter to stream.

plugins/plugin.go
```diff
@@ -16,6 +17,8 @@ const (
 	PluginTypeSink      PluginType = "sink"
 )
 
+type PushFunc func(models.Record)
```
ravisuhag, line 20: @StewartJingga I think this can still be kept. 
```
type Emitter func(models.Record)

Extract(ctx context.Context, emit plugin.Emitter) (err error)

emit()

**Review by ravisuhag — CHANGES_REQUESTED** [review]

File: agent/stream.go [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+type streamMiddleware func(src models.Record) (dst models.Record, err error)
+type subscription struct {
+	callback  func([]models.Record) error
+	channel   chan models.Record
+	batchSize int
+}
+
+type stream struct {
+	middlewares []streamMiddleware
+	listeners   []*subscription
```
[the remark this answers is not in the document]
StewartJingga: updated 🙏 , missed it before.

plugins/plugin.go
StewartJingga: updated to `plugin.Emit` 🙏

agent/batch.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"errors"
+
+	"github.com/odpf/meteor/models"
+)
+
+type batch struct {
+	data []models.Record
+	size int
```
kushsharma, line 11: This doesn't really reflect what it does. We should call it `capacity` I think.

agent/stream.go
```diff
@@ -0,0 +1,129 @@
...
+type subscriber struct {
+	callback  func([]models.Record) error
+	channel   chan models.Record
+	batchSize int
+}
+
+type stream struct {
+	middlewares []streamMiddleware
+	subscribers []*subscriber
+	hasClosed   bool
```
kushsharma, line 20: maybe `isClosed`? or just `closed`?

agent/stream.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	wg.Wait()
+
+	// there could be error when emitting
+	if s.err != nil {
+		return s.err
+	}
+
+	return nil
```
kushsharma, line 76: Can't we do this?
```suggestion
	return s.err
```

agent/stream.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// push() will run the record through all the registered middleware
+// and emit the record to all registerd subscribers.
```
kushsharma, line 80: typo `registerd`

plugins/extractors/bigtable/bigtable.go
```diff
@@ -114,9 +126,9 @@ func (e *Extractor) getTablesInfo(ctx context.Context, instances []string, proje
 					return
 				}
 				familyInfoBytes, _ := json.Marshal(tableInfo.FamilyInfos)
-				results = append(results, assets.Table{
+				emit(models.NewRecord(&assets.Table{
```
kushsharma, line 129: I thought we were against this before? Using pointer to pass data instead of struct? What made us change?

kushsharma (commented):
The new design looks great, just wanted to point out we are currently not handling graceful shutdown of the whole stream right?

plugins/extractors/bigtable/bigtable.go
StewartJingga: in most cases, I want it to be a struct instead of pointer.

But since we are using proto as the model, the linter points out that we are not supposed to pass proto by value because it will copy `sync.Mutex` from `protoimpl.MessageState`.

Since it is almost like a rule in using proto then I guess we kind of have to use pointer instead of value.

agent/stream.go
StewartJingga: yeah totally can do that, there were a lot of modification before, missed that changes.

agent/batch.go
StewartJingga: oh isnt batch size is a common word in this use case?

kushsharma: But this size variable doesn't really reflect our current batch size, size is reflected using `len(data)`, this variable is currently used to check if its full or not, and its more of a capacity of this batch not size.

StewartJingga:
> The new design looks great, just wanted to point out we are currently not handling graceful shutdown of the whole stream right?

@kushsharma  you mean in case of panic? yeah we could add that.

plugins/extractors/bigtable/bigtable.go
kushsharma: Nice, also this will also remove the need for reflection hack from our sinks, right?

kushsharma:
> > The new design looks great, just wanted to point out we are currently not handling graceful shutdown of the whole stream right?
> 
> @kushsharma you mean in case of panic? yeah we could add that.

Umm, more of a SIGTERM like pressing Ctrl+C or something.

StewartJingga:
> > > The new design looks great, just wanted to point out we are currently not handling graceful shutdown of the whole stream right?
> > 
> > 
> > @kushsharma you mean in case of panic? yeah we could add that.
> 
> Umm, more of a SIGTERM like pressing Ctrl+C or something.

ohhh yeah true, in that case it even needs to be handled from the root command.
right now it will just force shut down.

agent/batch.go
StewartJingga: oh truee! yeah that makes a lot of sense. updating it.

kushsharma:
Yeah, so the sequence of commands will go by first cancelling emitter, letting all sinks sink whatever is queued and then shut it down.

agent/agent.go
```diff
@@ -168,52 +153,79 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
...
+	err = extractor.Init(ctx, sr.Config)
+	if err != nil {
+		err = errors.Wrapf(err, "could not initiate extractor \"%s\"", sr.Type)
 		return
 	}
 
+	runFn = func() (err error) {
+		err = extractor.Extract(ctx, str.push)
+		if err != nil {
+			err = errors.Wrapf(err, "error running extractor \"%s\"", sr.Type)
```
kushsharma, line 171: We also need to cancel this run if extractor fails I guess? This should not be a soft fail?

kushsharma: Extractor and Processor errors should be hard fail I think, only sinks can soft fail.

StewartJingga: > We also need to cancel this run if extractor fails I guess? This should not be a soft fail?

this error will be escalated to https://github.com/odpf/meteor/pull/213/files?file-filters%5B%5D=.go#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR129

and it will close the stream immediately

StewartJingga: > Extractor and Processor errors should be hard fail I think, only sinks can soft fail.

right now everything is hard fail, but having sinks soft fail would be beneficial in multiple sinks case 🤔 .

will definitely consider that changes, but probably outside of this scope for now.

StewartJingga:
> Yeah, so the sequence of commands will go by first cancelling emitter, letting all sinks sink whatever is queued and then shut it down.

yup, basically we just need to cancel the emitter/stream, the rest will be handled gracefully by default, meaning that all sinks will finish consuming all pushed record then stream is done.

will do this in separate PR 🙏

agent/agent.go
kushsharma: Got it, missed that `stream.Close()` while handling error. Then this should be fine.

plugins/extractors/bigtable/bigtable.go
StewartJingga: created an issue https://github.com/odpf/meteor/issues/216 to fix kafka sink (where we did the reflection hack).
There was another problem and planning to fix it along with removing the hack there.

PR #221, opened 2021-09-17
fix: validate recipe name and plugin in gen

### LINKED ISSUE #207: Recipe generator returns invalid recipe when no plugin specified
When no plugin is specified in recipe generation, meteor still generates placeholders config. Which results in the recipe failing. 

```
processor:
```

ravisuhag:
closes #207


StewartJingga:
@ravisuhag i wonder if recipe generator can be handled in the recipe packages, so it would be like `recipe.Build()`.

But it seems like it would mess recipe package dependencies.

ravisuhag:
> @ravisuhag i wonder if recipe generator can be handled in the recipe packages, so it would be like `recipe.Build()`.
> 
> But it seems like it would mess recipe package dependencies.

Had the same thought and same concern. :) So kept it separate. I will add a note to consider this future if it makes sense.

StewartJingga (approved):
LGTM, but for handling indentation in the template, I was thinking to utilise [go template pipeline feature](https://pkg.go.dev/text/template#hdr-Pipelines).

so it would be like this
```
config:
{{ $value | indent 6 }}
```
similar to how helm handles it.

tried to do it on local but had to change a few things in all the plugins to achieve that, we can treat it as a good to have for now.

PR #222, opened 2021-09-17
feat: add superset extractor

scortier:
## Deliverables

- [x] add unit tests ( _TODO: Create user in superset_ )
- [x] add extractor
- [x] add README.md in ```plugins/extractors/superset```, defining output
- [x] register your extractor ```plugins/extractors/populate.go```
- [x] add extractor the extractor list in ```docs/reference/extractor.md```

## Outputs

| Field | Sample Value |
| :---- | :---- |
| `resource.urn` | `superset.dashboard_name` |
| `resource.name` | `dashboard_name` |
| `resource.service` | `superset` |
| `resource.url` | `dashboard_url` |

### Chart

| Field | Sample Value |
| :---- | :---- |
| `name` | `chart_name` |
| `dashboard_source` | `superset` |
| `description` | `chart_description` |
| `url` | `chart_url` |
| `datasource` | `chart_datasource` |
| `dashboard_urn` | `dashboard:dashboard_id` |

### DIFF
--- a/plugins/extractors/superset/superset.go
+++ b/plugins/extractors/superset/superset.go
@@ -1,0 +1,234 @@
+package superset
+
+import (
+	"bytes"
+	"context"
+	_ "embed" // used to print the embedded assets
+	"encoding/json"
+	"fmt"
+	"github.com/odpf/meteor/models"
+	"github.com/odpf/meteor/models/odpf/assets"
+	"github.com/odpf/meteor/models/odpf/assets/common"
+	"github.com/odpf/meteor/plugins"
+	"github.com/odpf/meteor/registry"
+	"github.com/odpf/meteor/utils"
+	"github.com/odpf/salt/log"
+	"github.com/pkg/errors"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+//go:embed README.md
+var summary string
+
+// Config hold the set of configuration for the superset extractor
+type Config struct {
+	Username string `mapstructure:"username" validate:"required"`
+	Password string `mapstructure:"password" validate:"required"`
+	Host     string `mapstructure:"host" validate:"required"`
+	Provider string `mapstructure:"provider" validate:"required"`
+}
+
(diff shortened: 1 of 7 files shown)


plugins/extractors/superset/typeDef.go
```diff
@@ -1,18 +1,48 @@
 package superset
 
 type Dashboard struct {
-	ID          int `json:"id"`
```
ravisuhag: @scortier We already dashboard model in proton. We should use that.

scortier: Yes will use that format only , can we add more model top over the prescribed format ?

ravisuhag: Do you mean more fields?

scortier: Yes

StewartJingga: @ravisuhag this model is just to catch superset model when fetching from its API.

This Dashboard model will then be mapped to our meteor model before sending to our stream.

StewartJingga: or maybe we can rename it to `SupersetModel` to avoid confusion.

plugins/extractors/superset/superset.go
```diff
@@ -0,0 +1,236 @@
...
+
+// buildDashboard builds a dashboard from superset server
+func (e *Extractor) buildDashboard(id int) (data *assets.Dashboard, err error) {
+	var dashboard Dashboard
+	chart, err := e.getChartsList(id)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get chart list")
+	}
+
+	var tempCharts []*assets.Chart
```
StewartJingga, line 117: can we move this mapping to `getChartsList` as well.
so this `getChartsList` will `assets.Chart` instead of superset's chart.

StewartJingga (changes requested):

StewartJingga (approved):
LGTM

PR #225, opened 2021-09-19
feat(metric): add data count

GrayFlash:


agent/agent.go
```diff
@@ -101,6 +101,7 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 		ctx         = context.Background()
 		getDuration = r.startDuration()
 		stream      = newStream()
+		dataCount   = 0
```
ravisuhag, line 104: @GrayFlash Let's call it `recordCount` ?

agent/monitor.go
```diff
@@ -8,13 +8,13 @@ import (
 
 // Monitor is the interface for monitoring the agent.
 type Monitor interface {
-	RecordRun(recipe recipe.Recipe, durationInMs int, success bool)
+	RecordRun(recipe recipe.Recipe, durationInMs, docCount int, success bool)
```
ravisuhag, line 11: How about we call this `recordCount`?

agent/agent.go
GrayFlash: Sure makes sense, will add it to the report table in `run.go` as the column named `Records` right??

metrics/statsd.go
```diff
@@ -29,35 +31,41 @@ func NewStatsdMonitor(client statsdClient, prefix string) *StatsdMonitor {
...
-		m.createMetricName(runDurationMetricName, recipe, success),
-		int64(duration),
+		m.createMetricName(runDurationMetricName, run.Recipe, run.Success, run.RecordCount),
+		int64(run.DurationInMs),
 	)
 	m.client.Increment(
-		m.createMetricName(runMetricName, recipe, success),
+		m.createMetricName(runMetricName, run.Recipe, run.Success, run.RecordCount),
+	)
+	m.client.IncrementByValue(
```
StewartJingga, line 42: i'm still thinking if we need another metric type for data count.
at first im thinking data count is just a metadata in a metric so we can put it as a tag.

let me try it out first if it we can actually get something out of this new metric.

PR #232, opened 2021-09-23
fix: catch panics to avoid breaking valid recipes

GrayFlash:


agent/agent.go
```diff
@@ -167,6 +168,7 @@ func (r *Agent) setupExtractor(ctx context.Context, sr recipe.SourceRecipe, str
 	}
 
 	runFn = func() (err error) {
+		defer recoverFunc()
```
StewartJingga, line 171: I wonder if we could handle the panic in the stream.go instead.

we could handle it on stream.broadcast and stream.push.

wdyt?

GrayFlash: Not sure how that will work, cause the runtime error is returned for the extractors when we get the returnFn also and not just the push and broadcast actually. Tried it earlier actually didn’t seem to work by my logic.

PR #235, opened 2021-09-24
feat: add couchdb extractor

GrayFlash:


plugins/extractors/couchdb/couchdb.go
```diff
@@ -0,0 +1,208 @@
...
+	Password string `mapstructure:"password" validate:"required"`
+	Host     string `mapstructure:"host" validate:"required"`
+}
+
+var sampleConfig = `
+ host: localhost:5984
+ user_id: admin
+ password: couchdb`
+
+// Extractor manages the extraction of data from MySQL
```
burntcarrot, line 42: Sorry for sounding picky, but I noticed that the comments still reflect `MySQL` as opposed to `CouchDB` while I was browsing the PR.

Similar occurrence at line 93.

GrayFlash: yup, missed changing that

plugins/extractors/couchdb/couchdb.go
```diff
@@ -0,0 +1,208 @@
...
+		return
+	}
+
+	return
+}
+
+// Extract extracts the data from the CouchDB server
+// and collected through the out channel
+func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error) {
+	defer e.client.Close(context.TODO())
```
StewartJingga, line 96: just use context from the argument instead of TODO. Applies to all occurrences.

StewartJingga (changes requested):

plugins/extractors/couchdb/couchdb.go
```diff
@@ -43,6 +43,7 @@ var sampleConfig = `
 type Extractor struct {
 	client      *kivik.Client
 	db          *kivik.DB
+	ctx         context.Context
```
StewartJingga, line 46: can we get by not storing context as internal state?

it is okay to keep passing context

GrayFlash: okay

StewartJingga (approved):

PR #238, opened 2021-09-27
feat: improve error handling

scortier:


agent/stream.go
```diff
@@ -78,11 +80,11 @@ func (s *stream) broadcast() error {
 }
 
 // push() will run the record through all the registered middleware
-// and emit the record to all registerd subscribers.
+// and emit the record to all registered subscribers.
 func (s *stream) push(data models.Record) {
 	data, err := s.runMiddlewares(data)
 	if err != nil {
-		s.err = errors.Wrap(err, "emitter: error running middleware")
+		s.err = errors.Wrapf(err, "emitter: error running middleware")
```
StewartJingga, line 87: we use `Wrapf` if we want to format our error message.

generator/recipe.go
```diff
@@ -30,43 +31,39 @@ func Recipe(name string, source string, sinks []string, processors []string) (er
 
 	if source != "" {
 		tem.Source = make(map[string]string)
-		sinfo, err := registry.Extractors.Info(source)
+		sourceInfo, err := registry.Extractors.Info(source)
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "failed to provide extractor information")
```
StewartJingga, line 36: same here, we use `Wrapf` if there is string formatting, we can use `Wrap` instead.

metrics/statsd.go
```diff
@@ -64,11 +65,11 @@ type statsdClient interface {
 func NewStatsdClient(statsdAddress string) (c *statsd.StatsdClient, err error) {
 	statsdHost, statsdPortStr, err := net.SplitHostPort(statsdAddress)
 	if err != nil {
-		return
+		return nil, errors.Wrapf(err, "failed to split a network address")
 	}
 	statsdPort, err := strconv.Atoi(statsdPortStr)
 	if err != nil {
-		return
+		return nil, errors.Wrapf(err, "failed to convert port type")
```
StewartJingga, line 72: to handle this kind of code, it's better to do it like below
```
if err != nil {
        err = errors.Wrap(err, "failed to split a network address")
	return
}
```

StewartJingga (changes requested):
Love the error wrapping but needs some changes on these two things:

1. Use `Wrap` instead of `Wrapf` if there is no string formatting.
2. For case where return value is already defined you can wrap error like below
```
if err != nil {
    err = errors.Wrap(err, "wrapping error")
    return
}
```

StewartJingga (changes requested):

agent/agent.go
```diff
@@ -145,14 +150,14 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
...
-			run.Error = err
+			run.Error = errors.Wrap(err, "failed to run extractor")
 		}
 	}()
 
 	// start listening.
 	// this process is blocking
 	if err := stream.broadcast(); err != nil {
-		run.Error = err
+		run.Error = errors.Wrap(err, "failed to stream")
```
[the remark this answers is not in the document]
scortier: sounds good 👍

StewartJingga (approved):
LGTM

PR #241, opened 2021-09-28
feat(bigquery): make preview configurable

StewartJingga:
- new `disable_preview` config on bigquery extractor

### DIFF
--- a/plugins/extractors/bigquery/bigquery.go
+++ b/plugins/extractors/bigquery/bigquery.go
@@ -233,15 +233,15 @@ func (e *Extractor) buildColumn(ctx context.Context, field *bigquery.FieldSchema
 func (e *Extractor) buildPreview(ctx context.Context, t *bigquery.Table) (fields []interface{}, preview []interface{}, err error) {
 	fields = []interface{}{}  // list of column names
 	preview = []interface{}{} // rows of column values
-	previewTotalRows := e.config.TotalPreviewRows
-	if previewTotalRows == 0 {
+	totalPreviewRows := e.config.TotalPreviewRows
+	if totalPreviewRows == 0 {
 		return
 	}
 
 	rows := []interface{}{}
 	totalRows := 0
 	ri := t.Read(ctx)
-	for totalRows < previewTotalRows {
+	for totalRows < totalPreviewRows {
 		var row []bigquery.Value
 		err = ri.Next(&row)
 		if err == iterator.Done {
--- a/plugins/extractors/bigquery/bigquery.go
+++ b/plugins/extractors/bigquery/bigquery.go
@@ -233,15 +233,14 @@ func (e *Extractor) buildColumn(ctx context.Context, field *bigquery.FieldSchema
 func (e *Extractor) buildPreview(ctx context.Context, t *bigquery.Table) (fields []interface{}, preview []interface{}, err error) {
 	fields = []interface{}{}  // list of column names
 	preview = []interface{}{} // rows of column values
-	maxPreviewRows := e.config.MaxPreviewRows
-	if maxPreviewRows == 0 {
+	if e.config.MaxPreviewRows == 0 {
 		return
 	}
 
 	rows := []interface{}{}
 	totalRows := 0
 	ri := t.Read(ctx)
-	for totalRows < maxPreviewRows {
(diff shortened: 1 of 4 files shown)


rohilsurana:
One thought I had was to let preview have its own config struct. I noticed that there is a var `previewTotalRows`. If in future we could have a need to expose it as a config and more such stuff can be introduced, we can as of now start with a struct for `PreviewConfigs` and have `disable` and `totalRows` as its fields.

Functionally the PR looks good to me.

StewartJingga:
> One thought I had was to let preview have its own config struct. I noticed that there is a var `previewTotalRows`. If in future we could have a need to expose it as a config and more such stuff can be introduced, we can as of now start with a struct for `PreviewConfigs` and have `disable` and `totalRows` as its fields.
> 
> Functionally the PR looks good to me.

Makes sense, but another thought is, should we just use `TotalPreviewRows`? we can put 0 to disable preview fetching and will default to 30 or other numbers.

Do you think it's better or do you see limitiations with it?

rohilsurana:
I think that could work as well, so we put a check on the `TotalPreviewRows` to return if its equal or less than 0, to avoid even triggering the BQ query.

I think we can start with this. It makes it implicit but should be fine.

StewartJingga:
@rohilsurana updated [here](https://github.com/odpf/meteor/pull/241/commits/f4e33a539a4d3c2b29538567fceb531f4c9f6801). 🙏

plugins/extractors/bigquery/bigquery.go
```diff
@@ -235,6 +233,10 @@ func (e *Extractor) buildColumn(ctx context.Context, field *bigquery.FieldSchema
 func (e *Extractor) buildPreview(ctx context.Context, t *bigquery.Table) (fields []interface{}, preview []interface{}, err error) {
 	fields = []interface{}{}  // list of column names
 	preview = []interface{}{} // rows of column values
+	previewTotalRows := e.config.TotalPreviewRows
```
rohilsurana, line 236: Lets use same variable naming - 
`TotalPreviewRows` and `totalPreviewRows` 
or
`PreviewTotalRows` and `previewTotalRows`.

StewartJingga: oh ok sure, was planning to not changing much lines. updating.

StewartJingga: updated

kushsharma: Or why total? how about `PreviewRowCount`

StewartJingga: hmm to be more exact, we can even use `MaxPreviewRows`, because there can be case where the total rows in the table is less than the row count config. wdyt? @kushsharma @rohilsurana

rohilsurana: `MaxPreviewRows` sounds good. 👍

StewartJingga: updated 🙏

rohilsurana (approved):

StewartJingga:
@rohilsurana  actually we need another flag for disable, so we can do like what you propose using nested config [here](https://github.com/odpf/meteor/pull/241#issuecomment-929836447).

reason is, when I put `0` as the config, it will be treated as an empty value and set it to its default value (`30` atm). So we would never be able to get `0` as the `MaxPreviewRows`.

rohilsurana:
That sounds like a bug of how we are loading the values into config struct, do you maybe want to address it there?
Is this happening using the `odpf/salt/config` package?

rohilsurana:
Just checked we are setting the defaults here - https://github.com/odpf/meteor/blob/dd15e2532141bd5fb7d7d0bd591edc4e66266132/utils/config.go#L22

I think we should load the defaults first into the struct and then the configs from configMap.
Is there any specific reason that I am missing to load defaults later?

StewartJingga:
> I think we should load the defaults first into the struct and then the configs from configMap.
Is there any specific reason that I am missing to load defaults later?

@rohilsurana oh no there's specific reasons, was just a wrong order and behaviour. Updated the PR with the fix 🙏 .

rohilsurana (approved):

PR #243, opened 2021-10-01
Add integration tests to test run recipes

scortier:


ravisuhag:
@scortier Let's move the e2e tests to the root test folder only. `test/e2e`

StewartJingga:
@ravisuhag right now `test` folder is used for test utilities, should we somehow restructure it?

the perks of having e2e to be it's own package is we can isolate the test easier.

ravisuhag:
> @ravisuhag right now `test` folder is used for test utilities, should we somehow restructure it?
> 
> the perks of having e2e to be it's own package is we can isolate the test easier.

It will be still isolated package inside test. `test/e2e`

StewartJingga:
we could do that as far as isolation goes.

but how would you define test package/folder now?
doing so will make test package to be
- a test utilities package required/depended by other packages
- a folder containing meteor tests (e2e, etc)

or if you want e2e to be inside `test` folder, maybe we can have it like this:
```bash
├── test
│   ├── utils
│   │   ├── dockertest.go
│   │   ├── logger.go
│   ├── mocks
│   │   ├── plugin.go
│   ├── e2e
│   │   ├── **/*_test.go
```

wdyt?

ravisuhag:
> we could do that as far as isolation goes.
> 
> but how would you define test package/folder now? doing so will make test package to be
> 
> * a test utilities package required/depended by other packages
> * a folder containing meteor tests (e2e, etc)
> 
> or if you want e2e to be inside `test` folder, maybe we can have it like this:
> 
> ```shell
> ├── test
> │   ├── utils
> │   │   ├── dockertest.go
> │   │   ├── logger.go
> │   ├── mocks
> │   │   ├── plugin.go
> │   ├── e2e
> │   │   ├── **/*_test.go
> ```
> 
> wdyt?

@scortier Yes, we can go with this proposed approach.

PR #244, opened 2021-10-02
feat: use go-template pipeline to handle indentation in recipe generated

GrayFlash:


GrayFlash:
Result:

image

Added a nindent function that helps you add a new line and then perform indentation, but can be removed since didn't use it anywhere.

generator/recipe.yaml
```diff
@@ -2,14 +2,14 @@ name: {{.Name}}
 source:
 {{- range $key, $value := .Source }}
   type: {{$key}}
-  config: {{$value}}
+  config: {{$value | indent 4}}
```
StewartJingga, line 5: this produces 5 indentations when generating bigquery extractor. This is not due to the logic of the template or the indent function, but it is because the `sampleConfig` in `bigquery` [extractor](https://github.com/odpf/meteor/blob/main/plugins/extractors/bigquery/bigquery.go#L40) has a space before each config.

can you help to make sure all of our sample configs does not have interfering whitespace on its own except the first new line?

StewartJingga (changes requested):
left a few small comments.

StewartJingga:
> Added a nindent function that helps you add a new line and then perform indentation, but can be removed since didn't use it anywhere.

`nindent` is actually pretty common usage and in most cases i'd be happy to have it ready.
but in our case, this templating is pretty internal to meteor contributor, so chances are it's wont be used unless we change our template, or our plugins` sample config value.

and about our sample config value, we need to have a standard for them (about newline) or it would be inconsistent for each plugins. currently we have newline before defining sample config in our plugins, which is helping for readability. This way newline is added by default via sample config and `nindent` is not needed.

other way to solve sample config consistency is by having `trim` function before indent to remove newlines, but I think we dont need to go there yet, this template generator is not user facing as in user wont touch the template itself.

as long as `meteor gen recipe` can produce consistent value then it should be okay, wdyt?

StewartJingga (approved):
LGTM

PR #245, opened 2021-10-02
feat(agent): add retry mechanism

StewartJingga:
- [x] move app config from agent to config package
- [x] create agent config
- [x] add retry mechanism with exponential backoff strategy
- [x] sink can be retried
- [x] retry interval and times can be configured
- [x] update Columbus sink to retry on 5** status code

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -240,22 +240,28 @@ func (r *Agent) setupSink(ctx context.Context, sr recipe.SinkRecipe, stream *str
 		return
 	}
 
-	stream.subscribe(func(records []models.Record) (err error) {
-		err = retryIfNeeded(
+	stream.subscribe(func(records []models.Record) error {
+		err := retryIfNeeded(
 			func() error {
 				return sink.Sink(ctx, records)
 			},
 			r.retryTimes,
 			r.retryInitialInterval,
 			func(e error, d time.Duration) {
-				r.logger.Info("retrying sink", "duration", d, "sink", sr.Name, "error", e.Error())
+				r.logger.Info(
+					fmt.Sprintf("retrying sink in %d", d),
+					"sink", sr.Name,
+					"error", e.Error())
 			},
 		)
+		// error (after exhausted retries) will just be skipped and logged
 		if err != nil {
-			return errors.Wrapf(err, "error running sink \"%s\"", sr.Name)
+			r.logger.Error("error running sink", "sink", sr.Name, "error", err.Error())
 		}
 
-		return
+		// TODO: create a new error to signal stopping stream.
+		// returning nil so stream wont stop.
+		return nil
 	}, defaultBatchSize)
 
 	return
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -25,7 +25,7 @@ type Agent struct {
 	sinkFactory          *registry.SinkFactory
 	monitor              Monitor
 	logger               log.Logger
-	retryTimes           int
+	maxRetries           int
 	retryInitialInterval time.Duration
 }
 
@@ -42,7 +42,7 @@ func NewAgent(config Config) *Agent {
 		sinkFactory:          config.SinkFactory,
 		monitor:              mt,
 		logger:               config.Logger,
-		retryTimes:           config.RetryTimes,
+		maxRetries:           config.MaxRetries,
 		retryInitialInterval: config.RetryInitialInterval,
 	}
 }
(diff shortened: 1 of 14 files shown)


agent/agent.go
```diff
@@ -20,24 +20,30 @@ const (
...
-	processorFactory *registry.ProcessorFactory
-	sinkFactory      *registry.SinkFactory
-	monitor          Monitor
-	logger           log.Logger
+	extractorFactory     *registry.ExtractorFactory
+	processorFactory     *registry.ProcessorFactory
+	sinkFactory          *registry.SinkFactory
+	monitor              Monitor
+	logger               log.Logger
+	retryTimes           int
```
ravisuhag, line 28: @StewartJingga If this represents no of retries. Then we can call it `maxRetries` WDYT?

agent/retry.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/cenkalti/backoff/v4"
+	"github.com/odpf/meteor/plugins"
+)
+
+const (
+	defaultRetryTimes      = 5
+	defaultInitialInterval = 5 * time.Second
+)
+
+func retryIfNeeded(operation func() error, retryTimes int, initialInterval time.Duration, notify func(e error, d time.Duration)) error {
```
ravisuhag, line 16: @StewartJingga How about we make retry interface to something like. 
```
attempt := retrier.new(interval, maxRetries)
attempt.make(fn)
```
We can find better names though. Wdyt?

StewartJingga:
@ravisuhag added a new update on sink:
1. default behaviour change so that run will not end if one record fails to be sunk (after retries are exhausted)
2. error will still be logged

next update would be to have an error for sink to return to signal the agent that recipe run needs to end.

agent/retry.go
StewartJingga: @ravisuhag  was trying to make it into a struct before.
but I was waiting for it to be more complex to justify it. wdyt?

agent/agent.go
```diff
@@ -240,22 +240,28 @@ func (r *Agent) setupSink(ctx context.Context, sr recipe.SinkRecipe, stream *str
...
+				r.logger.Info(
+					fmt.Sprintf("retrying sink in %d", d),
+					"sink", sr.Name,
+					"error", e.Error())
 			},
 		)
+		// error (after exhausted retries) will just be skipped and logged
 		if err != nil {
-			return errors.Wrapf(err, "error running sink \"%s\"", sr.Name)
+			r.logger.Error("error running sink", "sink", sr.Name, "error", err.Error())
```
ravisuhag, line 259: @StewartJingga Shall we make this behavior of skipping record after retry exhaust configurable?

StewartJingga: sure we could, but default would be to skip. wdyt?

agent/retry.go
ravisuhag: Readability is a challenge right now. So I would suggest that we can do it now.

agent/agent.go
ravisuhag: Skip is data loss. So I would say that should not be default.

StewartJingga: it is a data loss, but should we stop the run? 🤔 

also for the configuration, do you prefer it on recipe level or on meteor level?

ravisuhag: Let's keep meteor level for now.

agent/agent.go
```diff
@@ -234,14 +241,29 @@ func (r *Agent) setupSink(ctx context.Context, sr recipe.SinkRecipe, stream *str
 		return
 	}
 
-	stream.subscribe(func(records []models.Record) (err error) {
-		err = sink.Sink(ctx, records)
+	retryNotification := func(e error, d time.Duration) {
```
ravisuhag, line 244: @StewartJingga I think we can avoid this abstraction right. Since this retrier is very specific to the agent anyway. and we do not plan to expose it outside. Wdyt?

StewartJingga: sorry which abstraction? the `retryNotification`?
oh was just moving it from creating it again and again as anonymous func to a reusable functions, also im able to name it via variable for clarity. But im not against changing it back to anonymous func if you want.

btw it is still inside the agent.go code so nothing is exposed there, sink does need to know about it, this is just how agent.go handling sink.

ravisuhag: Ok. Lets skip for now.

PR #264, opened 2021-10-29
feat(columbus): add "service" field to columbus payload

StewartJingga:


models/metadata.go
```diff
@@ -9,4 +9,5 @@ import (
 type Metadata interface {
 	GetResource() *common.Resource
 	GetProperties() *facets.Properties
+	GetLineage() *facets.Lineage
```
ravisuhag, line 12: @StewartJingga This would not be applicable to every resource right? Like user.

StewartJingga: yes, but right now it is the only way Columbus sink is able to fetch lineage from our varies models.

The other way is to typecasting for each types which im not sure is scalable.

StewartJingga: @ravisuhag another approach is to have another interface for lineage and typecheck for that interface in Columbus.
```
type HasLineage interface {
  GetLineage() *facets.Lineage
}
```

ravisuhag: I think let's create another interface for lineage check instead of expecting all models to have it.

ravisuhag (approved):
LGTM

PR #265, opened 2021-11-03
feat(metabase): fetch dashboards and charts upstreams

StewartJingga:


models/urn.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+import "fmt"
+
+func TableUrn(service, host, database, name string) string {
+	return fmt.Sprintf("%s::%s/%s/%s", service, host, database, name)
+}
+
+func DashboardUrn(service, host, id string) string {
```
mabdh, line 9: For `Urn`, I think better to make it capitalized `URN` like the standard case in golang. ``

[Ref](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

```suggestion
func DashboardURN(service, host, id string) string {
```

plugins/extractors/metabase/metabase.go
```diff
@@ -103,33 +93,25 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error)
 		// "ordered_cards" field
 		dashboard, err := e.buildDashboard(d)
 		if err != nil {
-			return errors.Wrap(err, "failed to fetch dashboard data")
+			e.logger.Error("failed to build dashboard with", "dashboard_id", d.ID, "err", err.Error())
+		} else {
```
mabdh, line 97: Any reason why we use `else` here?

Why not

```suggestion
		dashboard, err := e.buildDashboard(d)
		if err != nil {
			e.logger.Error("failed to build dashboard with", "dashboard_id", d.ID, "err", err.Error())
			return
		}
		emit(models.NewRecord(dashboard))
```

plugins/extractors/metabase/metabase.go
```diff
@@ -147,125 +129,206 @@ func (e *Extractor) buildDashboard(d Dashboard) (data *assets.Dashboard, err err
...
+			cluster = tableComps[0]
+		}
+	case "bigquery":
+		if compLength > 2 {
+			cluster = tableComps[0]
+			dbName = tableComps[1]
+		} else if compLength > 1 {
+			dbName = tableComps[0]
+		}
+	}
```
mabdh, line 323: what would be the default case?

plugins/extractors/metabase/metabase.go
StewartJingga: oh no major reason really, was just trying to avoid using `continue`.
let's use your suggestion, but instead of return we'll use `continue`.

mabdh: I see, either way is fine though, just curious.

plugins/extractors/metabase/metabase.go
StewartJingga: @mabdh  the default case is there is no modification on `cluster` and `dbName` variables.
Do you think we should use `if else` instead of `switch` for better clarity?

mabdh: I see, no need to I guess. We can keep it like this.

mabdh (approved):

plugins/extractors/metabase/metabase.go
StewartJingga: > I see, either way is fine though

I like yours better 👌 

@mabdh updated it here
https://github.com/odpf/meteor/pull/265/commits/eacca61190efa465dba5f3a2de77c0bce4721671

PR #268, opened 2021-11-06
feat: allow interactive selection of plugin in info command

GrayFlash:
Allows Interactive Selection on which plugins info is needed. For the given command:
```meteor info <plugin_type> <plugin_name>```

Deliverables:
- [x] add selection for the name of the plugin

### CODE REVIEW

**Comment by GrayFlash** [comment]
Wanted to know if we plan to allow user use the info command like the way they use to earlier.
for e.g. do we want the command `meteor info extractor postgres` to still work.?

**Comment by ravisuhag** [comment]
@GrayFlash Yes, lets keep that for now.

**Comment by GrayFlash** [comment]
I am actually stuck on the part to provide a choice for `<plugin_type>`, i.e., extractor, sink, or processor.

What we want is that if the user enters, `meteor info` he must get an option to select `<plugin_type>` and if he enters `meteor info sink` he gets to the sink directly with an option to select from the list of sinks in the registry. The issue is that we were getting the name of the plugin, i.e., Postgres as input from args[] but the plugin_type is a part of the command. So I am not sure how should I move with this.

One solution would be running `info` as a command and taking the value of both `<plugin_type> <plugin_name>` as an argument, which if empty prompts for selection from the option we provide.

Any thoughts around this??

**Comment by ravisuhag** [comment]
@GrayFlash We can change `Args: cobra.ExactArgs(1)` in info <plugin> command to max arg one. and check if no arg is there show the user list of plugins to select from.

**Comment by GrayFlash** [comment]
Yup @ravisuhag but how would that help us add a selection if suppose `extractor` is missing from command `meteor info extractor` command because currently, the command seems to be `meteor info <command>` and if am not sure on how to add a prompt if someone enters `meteor info`.

**Comment by GrayFlash** [comment]
Did the implementation as per the discussion with @StewartJingga I have tried handling the case where used enters data beyond info but the suppose the command is incorrect using `ValidArgs`.
For e.g.
`$ meteor info sank`
Was throwing no error and just stopping without a log, so I have tried handling it. Do let me know of changes in approach or code hygiene.

**Comment by mabdh** [comment]
@GrayFlash  With `cobra`, could we do something like?

- If `meteor info`, call interactive command
- If `meteor info <arg>`, call sub command

If possible, maybe we can avoid using `switch-case` and use sub command instead.

**Comment by GrayFlash** [comment]
> @GrayFlash With `cobra`, could we do something like?
> 
> * If `meteor info`, call interactive command
> * If `meteor info <arg>`, call sub command
> 
> If possible, maybe we can avoid using `switch-case` and use sub command instead.

@mabdh even I wanted to have something like this, but I am not sure if it can be done.

Because, since we are getting the `plugin_type` as an arg, we will need to map it anyways using a switch-case.

Earlier it was being passed as a sub-command, like `meteor info` was parent and `extractor/processor/sink` would complete it. But there I was not able to pass a case to give interactive options to users in case the sub-command is empty.

If there are any suggestions on it please let me know.

**Comment by mabdh** [comment]
> > @GrayFlash With `cobra`, could we do something like?
> > 
> > * If `meteor info`, call interactive command
> > * If `meteor info <arg>`, call sub command
> > 
> > If possible, maybe we can avoid using `switch-case` and use sub command instead.
> 
> @mabdh even I wanted to have something like this, but I am not sure if it can be done.
> 
> Because, since we are getting the `plugin_type` as an arg, we will need to map it anyways using a switch-case.
> 
> Earlier it was being passed as a sub-command, like `meteor info` was parent and `extractor/processor/sink` would complete it. But there I was not able to pass a case to give interactive options to users in case the sub-command is empty.
> 
> If there are any suggestions on it please let me know.

hmm alright

@GrayFlash how about, we override `cmd.SetUsageFunc(f func(*Command) error)` to do interactive options?

**Comment by ravisuhag** [comment]
@GrayFlash @mabdh I think we can also have interactive mode, only on subcommand. And let the user specify the type explicitly.  `meteor info` will provide description/usage about that they should be one of them sub-command. WDYT?

**Comment by GrayFlash** [comment]
> @GrayFlash @mabdh I think we can also have interactive mode, only on subcommand. And let the user specify the type explicitly. `meteor info` will provide description/usage about that they should be one of them sub-command. WDYT?

I think we should move with this and revert back to using plugin_type being defined explicitly as there are just 3 options as of now, but to make it interactive I have made changes that may get a bit redundant. Any will keep the plugin as an interactive option.

For this issue, we can move with it and I will revert back with the changes. I will later play around with what @mabdh suggested and see if there is some possibility for that without making code this much hardcoded.

**Review by mabdh — APPROVED** [review]

**Comment by GrayFlash** [comment]
> @GrayFlash @mabdh I think we can also have interactive mode, only on subcommand. And let the user specify the type explicitly. `meteor info` will provide description/usage about that they should be one of them sub-command. WDYT?

@mabdh @ravisuhag I should make these changes right??

**Comment by ravisuhag** [comment]
> > @GrayFlash @mabdh I think we can also have interactive mode, only on subcommand. And let the user specify the type explicitly. `meteor info` will provide description/usage about that they should be one of them sub-command. WDYT?
> 
> @mabdh @ravisuhag I should make these changes right??

Yes, lets make these changes.

**Comment by mabdh** [comment]
> > @GrayFlash @mabdh I think we can also have interactive mode, only on subcommand. And let the user specify the type explicitly. `meteor info` will provide description/usage about that they should be one of them sub-command. WDYT?
> 
> @mabdh @ravisuhag I should make these changes right??

yes yes please

**Comment by GrayFlash** [comment]
@mabdh @ravisuhag is the `main` branch working fine for you all??

I pulled main branch into mine for resolving merge conflict. And it throwing this error, when I try `./meteor ` command:

`ERROR: unable to find config file, loading from env and defaults: Config File "config" Not Found in "[/home/gray/codes/t/meteor]"`

Do you know what config file is needed??

**Comment by mabdh** [comment]
> @mabdh @ravisuhag is the `main` branch working fine for you all??
> 
> I pulled main branch into mine for resolving merge conflict. And it throwing this error, when I try `./meteor ` command:
> 
> `ERROR: unable to find config file, loading from env and defaults: Config File "config" Not Found in "[/home/gray/codes/t/meteor]"`
> 
> Do you know what config file is needed??



> @mabdh @ravisuhag is the `main` branch working fine for you all??
> 
> I pulled main branch into mine for resolving merge conflict. And it throwing this error, when I try `./meteor ` command:
> 
> `ERROR: unable to find config file, loading from env and defaults: Config File "config" Not Found in "[/home/gray/codes/t/meteor]"`
> 
> Do you know what config file is needed??

hi @GrayFlash, I have tested in my local and yes it returns error. Normally viper needs `config.yaml` file. But I think the config should automatically load from `mapstructure` default. Last time it worked for me, let me see what is missing.

any thought on this @StewartJingga ?

**Comment by StewartJingga** [comment]
Meteor does not handle config error from salt properly.
Here is the PR for the fix.
https://github.com/odpf/meteor/pull/304

**Comment by mabdh** [comment]
I think this changes https://github.com/odpf/salt/commit/60f6178cfd80cc80b4905e940a435501acb31125 caused the error above

**Comment by ravisuhag** [comment]
@GrayFlash Yes, you would need a config file at the root for now, Unless another pending PR is closed.

File: cmd/run.go [hunk 1]
```diff
@@ -94,7 +94,10 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 					row = append(row, cs.SuccessIcon(), run.Recipe.Name, cs.Grey(run.Recipe.Source.Type), cs.Greyf("%v ms", strconv.Itoa(run.DurationInMs)), cs.Greyf(strconv.Itoa(run.RecordCount)))
 				}
 				report = append(report, row)
-				bar.Add(1)
+				err := bar.Add(1)
```
GrayFlash, line 97: Added it here to fix the lint-ci

ravisuhag:
@GrayFlash Please see if you can find some time to close this.

GrayFlash:
> @GrayFlash Please see if you can find some time to close this.

Yes, @ravisuhag this one is up for review. The changes made were to get back on allowing interactive options for the name of plugin, and keeping the type something user mentions.

ravisuhag (approved):

PR #273, opened 2021-11-29
feat(optimus): use grpc to connect

StewartJingga:


plugins/extractors/optimus/client.go
```diff
@@ -33,25 +32,25 @@ func newClient() Client {
 
 type client struct {
 	pb.RuntimeServiceClient
+	conn *grpc.ClientConn
 }
 
 func (c *client) Connect(ctx context.Context, host string) (err error) {
 	dialTimeoutCtx, dialCancel := context.WithTimeout(ctx, time.Second*2)
```
mabdh, line 39: wouldn't be better if we centralize this config timeout`time.Second*2` into the config struct itself?, although it wouldn't be configurable from recipe.

StewartJingga: we could I guess, we can accept a string from recipe and then parse it to `time.Duration` using `time.ParseDuration()`.
https://pkg.go.dev/time#ParseDuration

wdyt?

mabdh: yes we could

PR #286, opened 2021-12-16
feat(extractor): populate email to ownership in optimus and tableau

mabdh:


plugins/extractors/optimus/optimus.go
```diff
@@ -149,8 +149,9 @@ func (e *Extractor) buildJob(ctx context.Context, jobSpec *pb.JobSpecification,
 		Ownership: &facets.Ownership{
 			Owners: []*facets.Owner{
 				{
-					Urn:  jobSpec.Owner,
-					Name: jobSpec.Owner,
+					Urn:   jobSpec.Owner,
+					Name:  jobSpec.Owner,
```
StewartJingga, line 153: Since we already have Email field and optimus does not store name, should we just skip Name? Wdyt? @mabdh

mabdh: yes, fair enough.

StewartJingga (approved):
LGTM

PR #290, opened 2022-01-03
feat: connect database with connection URL

scortier:
- Update the extractor config by replacing login credentials with the dynamic connection URL.

### DIFF
--- a/plugins/extractors/mssql/mssql.go
+++ b/plugins/extractors/mssql/mssql.go
@@ -33,8 +33,8 @@ var defaultDBList = []string{
 
 // Config holds the set of configuration for the extractor
 type Config struct {
-	UserID   string `mapstructure:"user_id" validate:"required"`
-	Password string `mapstructure:"password" validate:"required"`
+	UserID   string `mapstructure:"user_id" default:"root"`
+	Password string `mapstructure:"password"`
 	Host     string `mapstructure:"host" validate:"required"`
 }
 
--- a/plugins/extractors/mssql/mssql.go
+++ b/plugins/extractors/mssql/mssql.go
@@ -31,17 +31,13 @@ var defaultDBList = []string{
 	"tempdb",
 }
 
-// Config holds the set of configuration for the extractor
+// Config hold the connection URL for the extractor
 type Config struct {
-	UserID   string `mapstructure:"user_id" default:"root"`
-	Password string `mapstructure:"password"`
-	Host     string `mapstructure:"host" validate:"required"`
+	ConnectionURL string `mapstructure:"connection_url" validate:"required"`
 }
 
 var sampleConfig = `
-host: localhost:1433
-user_id: admin
-password: "1234"`
+connection_url: "sqlserver://admin:pass123@localhost:3306/"`
 
 // Extractor manages the extraction of data from the database
 type Extractor struct {
@@ -85,7 +81,7 @@ func (e *Extractor) Init(ctx context.Context, configMap map[string]interface{}) 
 	e.buildExcludedDBs()
 
 	// create client
-	if e.db, err = sql.Open("mssql", fmt.Sprintf("sqlserver://%s:%s@%s/", e.config.UserID, e.config.Password, e.config.Host)); err != nil {
+	if e.db, err = sql.Open("mssql", e.config.ConnectionURL); err != nil {
 		return errors.Wrap(err, "failed to create client")
 	}
 
--- a/plugins/extractors/mssql/mssql.go
+++ b/plugins/extractors/mssql/mssql.go
@@ -31,7 +31,7 @@ var defaultDBList = []string{
 	"tempdb",
 }
 
-// Config hold the connection URL for the extractor
+// Config holds the connection URL for the extractor
 type Config struct {
 	ConnectionURL string `mapstructure:"connection_url" validate:"required"`
 }
--- a/plugins/extractors/postgres/README.md
+++ b/plugins/extractors/postgres/README.md
@@ -6,7 +6,7 @@
 source:
   type: postgres
   config:
-    connection_url: postgres://admin:pass123@localhost:3306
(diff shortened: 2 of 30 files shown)


plugins/extractors/postgres/README.md
```diff
@@ -9,6 +9,8 @@ source:
     host: localhost:5432
     user_id: admin
```
mabdh, line 10: any reason why we don't make postgres to use connection url like the others?

StewartJingga: currently postgres extractor requires `host` and `database` value to be used as part of the `URN`.

If we are using connection URL, we would have to get the host and database value from the connection string.
It is totally possible but there could be different formats of connection string and we might need to handle each formats if we want to extract the host and database.

We could:
- change postgres to use connection string and handle connection string formats
- change postgres to use connection string and change how we generate urn so that it does not depend on host

mabdh: If that's the case, sooner or later the other DB also will have `host` as part of its URN right? or won't it happen?

scortier: If later other DB extractor requires `host` as a part of `URN` then either we have to change 
- how we generate `urn` to satisfy all connection string format 
- have to use username and password separately just like we use currently
wdyt ??

mabdh: as for now, I don't think that will be much issue. Just need to take a note that other DBs are still using URN `database_name.table_name`. The URN afaik is not final and could still be evolving.

In that case, can you please also update the resource URN in the output in this Postgres [README](https://github.com/odpf/meteor/blob/main/plugins/extractors/postgres/README.md)?

kushsharma: Even I think we should standardize as @mabdh pointed out, use the connection URL here as well. It's pretty easy to parse connection strings as the logic of parsing can stay in individual extractors right?

mabdh: Yes better to make it consistent so it is predictable. 

It is not that difficult to parse URI, we can use [net/url](https://pkg.go.dev/net/url#Parse) or [libpq](https://github.com/lib/pq/blob/master/url.go#L32).

plugins/extractors/postgres/postgres.go
```diff
@@ -181,9 +178,14 @@ func (e *Extractor) getTableMetadata(db *sql.DB, dbName string, tableName string
 		return result, nil
 	}
 
+	host, err := splitURL(e.config)
```
mabdh, line 181: Parsing the URL connection could be done earlier, even it is possible to be parsed once the Extract function is called since this `host` value would be the same for all the metadata in this context. It is better to fail fast so user could catch and fix the error earlier.

plugins/extractors/postgres/postgres.go
```diff
@@ -181,9 +178,14 @@ func (e *Extractor) getTableMetadata(db *sql.DB, dbName string, tableName string
...
+	if err != nil {
+		err = errors.Wrap(err, "failed to split host from connection string")
+		return
+	}
 	result = &assetsv1beta1.Table{
 		Resource: &commonv1beta1.Resource{
-			Urn:     models.TableURN("postgres", e.config.Host, dbName, tableName),
```
mabdh, line 186: We could also still have a `host` config but instead of exposing it and populated externally, we populate it internally with parse url.

plugins/extractors/postgres/postgres.go
```diff
@@ -230,10 +232,22 @@ func isNullable(value string) bool {
 
 // connection generates a connection string
 func connection(cfg Config, database string) (db *sql.DB, err error) {
-	connStr := fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=disable", cfg.UserID, cfg.Password, cfg.Host, database)
+	connStr := fmt.Sprintf("%s/%s?sslmode=disable", cfg.ConnectionURL, database)
 	return sql.Open("postgres", connStr)
 }
 
+// splitURL splits the connection string into the host
+func splitURL(cfg Config) (host string, err error) {
```
mabdh, line 240: The function does not really `split` the URL. It is actually just `fetching` host from Connection URL. I think the proper name would be something align with `GetHostFromConnectionURL`.

plugins/extractors/mssql/mssql.go
```diff
@@ -31,17 +31,13 @@ var defaultDBList = []string{
 	"tempdb",
 }
 
-// Config holds the set of configuration for the extractor
+// Config hold the connection URL for the extractor
```
mabdh, line 34:
```suggestion
// Config holds the connection URL for the extractor
```

plugins/extractors/postgres/README.md
```diff
@@ -6,25 +6,23 @@
 source:
   type: postgres
   config:
-    host: localhost:5432
-    user_id: admin
-    password: 1234
+    connection_url: postgres://admin:pass123@localhost:3306/testDB?sslmode=disable
+    database_name: testDB
```
mabdh, line 10: btw, is `database_name` here really needed?

scortier: Removing it, forgot to remove it after changes

plugins/extractors/postgres/postgres.go
```diff
@@ -27,25 +28,24 @@ var defaults = []string{"information_schema", "root", "postgres"}
...
 
 // Extractor manages the extraction of data from the extractor
 type Extractor struct {
-	logger log.Logger
-	config Config
-	db     *sql.DB
+	logger   log.Logger
+	config   Config
+	db       *sql.DB
+	userDB   string
```
scortier, line 44: Any thought, how we can remove `userDB` from `Extractor` struct and initialize `*sql.DB` directly.

mabdh: @scortier can you please give more context ?

mabdh (approved):

PR #291, opened 2022-01-05
fix: inconsistent test on agent module

scortier:


agent/agent_test.go
```diff
@@ -546,8 +546,8 @@ func TestRunnerRun(t *testing.T) {
 			Recipe:      validRecipe,
 			RecordCount: 1,
 			Success:     true,
-			// DurationInMs is 0 because of the process took below a millisecond, in production scenario it takes longer than that
-			DurationInMs: 0,
+			// DurationInMs is 1000 as an arbitary alloted duration time, in production scenario it takes longer than that
+			DurationInMs: 1000,
```
StewartJingga, line 550: let's store this `1000` as a variable e.g. `expectedDuration` instead of hardcoding it here and line 564 below

agent/agent_test.go
```diff
@@ -546,8 +546,8 @@ func TestRunnerRun(t *testing.T) {
 			Recipe:      validRecipe,
 			RecordCount: 1,
 			Success:     true,
-			// DurationInMs is 0 because of the process took below a millisecond, in production scenario it takes longer than that
-			DurationInMs: 0,
+			// DurationInMs is 1000 as an arbitary alloted duration time, in production scenario it takes longer than that
```
StewartJingga, line 549: we can also remove this comment since the variable name should be pretty descriptive

agent/config.go
```diff
@@ -16,4 +16,5 @@ type Config struct {
 	MaxRetries           int
 	RetryInitialInterval time.Duration
 	StopOnSinkError      bool
+	TimerFn              func() func() int
```
StewartJingga, line 19: let's create a type for this function e.g. TimerFn instead of using `func()func()int` everywhere

StewartJingga (changes requested):

agent/agent_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+var expectedDuration = 1000
```
StewartJingga, line 21: let's just put this expectedDuration variable inside the scope of the test that is using it only.

StewartJingga (changes requested):

agent/agent_test.go
scortier: The `expectedDuration` is also used on line 672 , so decided to put it on global scope.

Issue #256, opened 2021-10-20
bug: unrecorded metrics on failed plugin initialization


StewartJingga:
**Describe the bug**
Error on plugins' `Init` method does not record any metrics.

**To Reproduce**
1. Create a kafka recipe with unreachable broker
2. Setup statsd config in Meteor (e.g. `STATSD_ENABLED`) and any metrics visualisation tools (e.g. `Grafana`)
3. Run recipe
4. Confirm if run metrics are not recorded

**Expected behavior**
Metrics should be collected even on failed `Init` methods

**Screenshots**

**Desktop (please complete the following information):**
 - Version: `v0.1.11`

PR #292, opened 2022-01-05
fix: unrecorded metrics on failed plugin initialization

ishanarya0:
Fixes for the issue: https://github.com/odpf/meteor/issues/256

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -113,54 +113,81 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 		stream      = newStream()
 		recordCount = 0
 	)
+	var errs error
+
 	runExtractor, err := r.setupExtractor(ctx, recipe.Source, stream)
 	if err != nil {
-		run.Error = errors.Wrap(err, "failed to setup extractor")
-		return
+		err = errors.Wrap(err, "failed to setup extractor")
+		if errs == nil {
+			errs = errors.New("")
+		}
+		errs = errors.Wrap(errs, err.Error())
 	}
 
 	for _, pr := range recipe.Processors {
 		if err := r.setupProcessor(ctx, pr, stream); err != nil {
-			run.Error = errors.Wrap(err, "failed to setup processor")
-			return
+			err = errors.Wrap(err, "failed to setup processor")
+			if errs == nil {
+				errs = errors.New("")
+			}
+			errs = errors.Wrap(errs, err.Error())
 		}
 	}
 
 	for _, sr := range recipe.Sinks {
 		if err := r.setupSink(ctx, sr, stream); err != nil {
-			run.Error = errors.Wrap(err, "failed to setup sink")
-			return
+			err = errors.Wrap(err, "failed to setup sink")
+			if errs == nil {
+				errs = errors.New("")
+			}
+			errs = errors.Wrap(errs, err.Error())
 		}
 	}
 
-	// to gather total number of records extracted
-	stream.setMiddleware(func(src models.Record) (models.Record, error) {
-		recordCount++
-		return src, nil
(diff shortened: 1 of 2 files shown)


ishanarya0:
Kindly review this PR @StewartJingga @mabdh  @ravisuhag

agent/agent.go
```diff
@@ -113,54 +113,81 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
...
+		if err := stream.broadcast(); err != nil {
+			err = errors.Wrap(err, "failed to broadcast stream")
+			if errs == nil {
+				errs = errors.New("")
+			}
+			errs = errors.Wrap(errs, err.Error())
+		}
 	}
 
 	// code will reach here stream.Listen() is done.
```
mabdh, line 189: Instead of checking err = nil everywhere like this
```
if errs == nil {
  errs = errors.New("")
}
```
Could we just wrap these part of code where we gather the metrics in a function and use defer to call that function. In that case, the wrapped function will always be called just before return.

agent/agent_test.go
```diff
@@ -52,12 +52,14 @@ func TestRunnerRun(t *testing.T) {
 
 	t.Run("should return error if extractor could not be found", func(t *testing.T) {
 		proc := mocks.NewProcessor()
+		proc.On("Init", mockCtx, validRecipe.Processors[0].Config).Return(nil).Once()
```
StewartJingga, line 55: why do we need this? IIRC, if extractor could not be found, then processors and sinks Init won't be called.

StewartJingga (changes requested):

agent/agent_test.go
StewartJingga: Also on this [test](https://github.com/odpf/meteor/pull/292/files/8b706c97ac7c3a3ef5fb112653dbf7155d5269cf#diff-e12d66df9a328a0b4b4e61abbde9a65eb60a4c8a6a8ca9249a455f173007ac89R133), you need to assert that `RecordRun` are being called instead of plugins and sinks init methods.

ishanarya0: > why do we need this? IIRC, if extractor could not be found, then processors and sinks Init won't be called.

I changed the flow such that even though extractor initialization fails, it will try to initialization processor and sink. Though it will not try to extract further. The purpose of doing this was to capture all the failing initializations at once and report them to the user.

Do we want this change or shall we step back to the previous one?

ishanarya0: > Also on this [test](https://github.com/odpf/meteor/pull/292/files/8b706c97ac7c3a3ef5fb112653dbf7155d5269cf#diff-e12d66df9a328a0b4b4e61abbde9a65eb60a4c8a6a8ca9249a455f173007ac89R133), you need to assert that `RecordRun` are being called instead of plugins and sinks init methods.

Yes, I'll add that to the test.

agent/agent.go
```diff
@@ -113,6 +113,11 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 		stream      = newStream()
 		recordCount = 0
 	)
+
+	defer func() {
+		r.runRecordRun(run, getDuration, recipe.Name)
+	}()
```
mabdh, line 119: Wouldn't it be better written like this?
```suggestion
	defer r.runRecordRun(run, getDuration, recipe.Name)
```

ishanarya0: > Wouldn't it be better written like this?

The parameter values in the runRecordRun change their value, so in order to pass the latest value, I used this closure approach.

mabdh (approved):

StewartJingga (approved):
LGTM

PR #293, opened 2022-01-06
fix: upgrade optimus extractor rpc to use v1beta1 APIs

kushsharma:
Recently Optimus APIs have done a breaking change to version it starting from `v1beta1`. Optimus will not keep its APIs backward compatible at the moment so this is required for it to work moving forward.


StewartJingga (approved):
Code's looking good, will test it on staging first before merging

PR #294, opened 2022-01-07
feat: add oracle extractor

Srtpatil:
### Deliverables

- [x] add unit tests
- [x] add extractor
- [x] add README.md in plugins/extractors/oracle, defining output
- [x] register your extractor plugins/extractors/populate.go
- [x] add extractor the extractor list in docs/reference/extractor.md

### Related Issue
#95

### DIFF
--- a/plugins/extractors/oracle/README.md
+++ b/plugins/extractors/oracle/README.md
@@ -6,7 +6,7 @@
 source:
   type: oracle
   config:
-		connection_url: admin/1234@localhost:1521/xe
+    connection_url: admin/1234@localhost:1521/xe
 ```
 
 ## Inputs
--- a/plugins/extractors/oracle/README.md
+++ b/plugins/extractors/oracle/README.md
@@ -1,0 +1,45 @@
+# Oracle
+
+## Usage
+
+```yaml
+source:
+  type: oracle
+  config:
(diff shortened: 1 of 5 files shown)


plugins/extractors/oracle/README.md
```diff
@@ -0,0 +1,45 @@
+# Oracle
+
+## Usage
+
+```yaml
+source:
+  type: oracle
+  config:
+    host: localhost:1521
```
mabdh, line 9: is it possible to use connection string to connect DB? if yes I think better to do it like that so we have consistent way to connect DB for all of our DB extractors

this is related with https://github.com/odpf/meteor/pull/290

Srtpatil: Sure, We can add the connection to DB via a connection string.

mabdh (approved):

plugins/extractors/oracle/README.md
```diff
@@ -0,0 +1,39 @@
+# Oracle
+
+## Usage
+
+```yaml
+source:
+  type: oracle
+  config:
+		connection_url: admin/1234@localhost:1521/xe
```
mabdh, line 9: can we fix the tab in this line?

Issue #223, opened 2021-09-18
Report recipe run progress in realtime 


ravisuhag:
Meteor should provide recipe run progress in real-time when running multiple recipes. This data can be presented as a progress bar in CLI. Since meteor can also be used in deployment mode. Meteor run command can accept a flag for daemon mode which does not report this and only uses standard logging.

PR #299, opened 2022-01-14
feat: add progress bar

GrayFlash:
Adding a progress bar denoting the progress of `meteor run recipe` command.

image

Have used `github.com/cheggaaa/pb/` for the purpose. Let me know if yo have seen some other alternative based on plans around meteor!!

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -79,7 +79,6 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 				progressbar.OptionSetDescription("[cyan]running recipes [reset]"),
 				progressbar.OptionShowCount(),
 			)
-			//.Default(int64(len(recipes)))
 
 			// Run recipes and collect results
 			runs := runner.RunMultiple(recipes)
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/MakeNowJust/heredoc"
+	"github.com/cheggaaa/pb/v3"
 	"github.com/odpf/meteor/agent"
 	"github.com/odpf/meteor/config"
 	"github.com/odpf/meteor/metrics"
@@ -73,6 +74,7 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 			var failures = 0
 			report = append(report, []string{"Status", "Recipe", "Source", "Duration(ms)", "Records"})
 
+			bar := pb.StartNew(len(recipes))
 			// Run recipes and collect results
 			runs := runner.RunMultiple(recipes)
 			for _, run := range runs {
@@ -87,7 +89,9 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 					row = append(row, cs.SuccessIcon(), run.Recipe.Name, cs.Grey(run.Recipe.Source.Type), cs.Greyf("%v ms", strconv.Itoa(run.DurationInMs)), cs.Greyf(strconv.Itoa(run.RecordCount)))
 				}
 				report = append(report, row)
+				bar.Increment()
 			}
+			bar.Finish()
 
 			// Print the report
 			if failures > 0 {
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/MakeNowJust/heredoc"
-	"github.com/cheggaaa/pb/v3"
 	"github.com/odpf/meteor/agent"
 	"github.com/odpf/meteor/config"
 	"github.com/odpf/meteor/metrics"
@@ -16,6 +15,7 @@ import (
 	"github.com/odpf/salt/log"
 	"github.com/odpf/salt/printer"
 	"github.com/odpf/salt/term"
+	"github.com/schollz/progressbar/v3"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


ravisuhag:
@GrayFlash Check this also and see which one is better. 
https://github.com/schollz/progressbar

GrayFlash:
> @GrayFlash Check this also and see which one is better. https://github.com/schollz/progressbar

Hey, checked both. Both are good in terms of number of people using them, and https://github.com/cheggaaa/pb/ is good in terms of a minimal progress bar with not many modifications needed.

But I think we should use https://github.com/schollz/progressbar cause it is more customizable in terms of color, themes, etc.

wdyt?? @ravisuhag

ravisuhag:
> > @GrayFlash Check this also and see which one is better. https://github.com/schollz/progressbar
> 
> Hey, checked both. Both are good in terms of number of people using them, and https://github.com/cheggaaa/pb/ is good in terms of a minimal progress bar with not many modifications needed.
> 
> But I think we should use https://github.com/schollz/progressbar cause it is more customizable in terms of color, themes, etc.
> 
> wdyt?? @ravisuhag

Sounds good.

GrayFlash:
Screenshot from 2022-01-18 11-11-10

Hey @ravisuhag updated the proogress bar.

GrayFlash:
#223 mentions that "Meteor can be used in deployment mode. Meteor run command can accept a flag for daemon mode which does not report this and only uses standard logging." 

So what further changes are required for this complete issue??

ravisuhag:
> #223 mentions that "Meteor can be used in deployment mode. Meteor run command can accept a flag for daemon mode which does not report this and only uses standard logging."
> 
> So what further changes are required for this complete issue??

@GrayFlash Ideally we would have the `isCI` function in the salt term package and use that to disable the progress bar in CI.

StewartJingga:
> @GrayFlash Ideally we would have the isCI function in the salt term package and use that to disable the progress bar in CI.
> For now, we can add a flag to do it the non-interactive way which just disables the progress bar.
> 
> @StewartJingga WDYT?


+1, we could have a specific flag to disable this feature, like e.g. `-hide-progress-bar` to be more explicit, where if it is not defined will default to progress bar to be shown.

@GrayFlash @ravisuhag

cmd/run.go
```diff
@@ -73,6 +74,13 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 			var failures = 0
 			report = append(report, []string{"Status", "Recipe", "Source", "Duration(ms)", "Records"})
 
+			bar := progressbar.NewOptions(len(recipes),
+				progressbar.OptionEnableColorCodes(true),
+				progressbar.OptionSetDescription("[cyan]running recipes [reset]"),
+				progressbar.OptionShowCount(),
+			)
+			//.Default(int64(len(recipes)))
```
StewartJingga, line 82: don't forget to remove this line

StewartJingga (changes requested):

ravisuhag:
> > @GrayFlash Ideally we would have the isCI function in the salt term package and use that to disable the progress bar in CI.
> > For now, we can add a flag to do it the non-interactive way which just disables the progress bar.
> > @StewartJingga WDYT?
> 
> +1, we could have a specific flag to disable this feature, like e.g. `-hide-progress-bar` to be more explicit, where if it is not defined will default to progress bar to be shown.
> 
> @GrayFlash @ravisuhag

Let's call flag as `deamon` shortversion as `-d` ?

StewartJingga:
> Let's call flag as daemon shortversion as -d ?

And if the flag is there Meteor only outputs logs? @ravisuhag

ravisuhag:
> > Let's call flag as daemon shortversion as -d ?
> 
> 
> 
> And if the flag is there Meteor only outputs logs? @ravisuhag 

Yes

cmd/run.go
GrayFlash: oh sorry, forgot to save the file before commit.. 😅

kushsharma:
Optimus does it a little [differently](https://github.com/odpf/optimus/blob/main/cmd/progressbar.go), it checks if the output stream is a `tty`, if yes, only then shows the progress bar. It also supports taking an env variable `OPTIMUS_PROGRESS_INDICATOR` to disable it on demand.

Also, where are you streaming this progress bar? Is it `stdout` or `stderr`?

ravisuhag:
> Optimus does it a little [differently](https://github.com/odpf/optimus/blob/main/cmd/progressbar.go), it [checks](https://github.com/odpf/optimus/blob/main/cmd/commands.go#L207) if the output stream is a `tty`, if yes, only then shows the progress bar. It also supports taking an env variable `OPTIMUS_PROGRESS_INDICATOR` to disable it on demand.
> 
> Also, where are you streaming this progress bar? Is it `stdout` or `stderr`?

+1 Have added a tty check function is [salt](https://github.com/odpf/salt/blob/9aa38ffc5626572e756f1abba7f1485b6e3de1d1/term/term.go#L10). 

We should also abstract out the progress bar as a lib in salt to standardize it across projects.

ravisuhag:
@GrayFlash We can merge this for now. Later we will abstract out the progress bar in salt which also checks the TTY presence and use that. For now, it should be good.

kushsharma:
Did we fixed the output of progress bar to be sent to `stderr` instead of `stdout`? If we are piping the output to a different command, like `grep`, you wouldn't want to mix the normal application output with progress bar.

ravisuhag:
@kushsharma I am working on a PR to abstract that out in salt and we will use that here. PR coming soon. So merged this for now.

PR #300, opened 2022-01-18
fix: use cgofree oracle client

ravisuhag:


ravisuhag:
Hi @Srtpatil godror uses CGO which we are avoiding using in meteor for now. Can you check/test if this client is fine?

Srtpatil:
Sure I will take a look

ravisuhag:
HIgh-level tests are passing. Merging it to avoid breaking the pipeline. e2e testing for this can be picked up in parallel.

PR #304, opened 2022-01-25
fix(config): error when running without config file

StewartJingga:


config/config.go
```diff
@@ -19,8 +22,9 @@ func Load() (cfg Config, err error) {
 	err = config.
 		NewLoader(config.WithPath("./")).
 		Load(&cfg)
-	if err != nil {
-		return
+	if errors.As(err, &config.ConfigFileNotFoundError{}) {
+		log.Println(err)
```
mabdh, line 26: if we need to print the error, I think we can just use `info` or `warn`, if the config is not `required`.

scortier: We can use `info` from `salt/log` here

StewartJingga: agree! but at this point, logger is not yet initialised and we are left with logger from standard package. @mabdh

mabdh: @StewartJingga  yes since it is the initiation part I guess we can't use `salt/log`.
maybe could also simply just use `fmt.Println` like we print the error when failed to load the config https://github.com/odpf/meteor/blob/main/main.go#L28.

mabdh: another alternative is, can we do this error check in `main.go` instead? I think it would be better to use `fmt.Println` there.

mabdh: either way is fine I guess. although using `log` standard package is also fine.

PR #305, opened 2022-01-25
feat(metrics): update statsd metric to include extractor name

scortier:


mabdh (approved):

agent/run.go
```diff
@@ -16,9 +16,10 @@ const (
 
 // Run contains the json data
 type Run struct {
-	Recipe       recipe.Recipe `json:"recipe"`
-	Error        error         `json:"error"`
-	DurationInMs int           `json:"duration_in_ms"`
-	RecordCount  int           `json:"record_count"`
-	Success      bool          `json:"success"`
+	Extractor    recipe.SourceRecipe `json:"extractor"`
```
ravisuhag, line 19: @scortier Do we need to add extractor here or this information is already available In recipe? cc @StewartJingga

mabdh: @scortier oh yes we could just get extractor type from `Recipe.SourceRecipe.Type`

scortier: Yes , we can use it, but for explicit detailing of the solution, included it.
@StewartJingga wdyt ?

StewartJingga: @scortier  I think getting it from the current `Recipe` field would be better. Adding another extractor field might give the wrong idea that it is a different extractor from the one in the Recipe.

StewartJingga (approved):

PR #307, opened 2022-01-27
feat: log extracted record information

scortier:
### Extracted record preview : 

image

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -152,6 +152,7 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 	// to gather total number of records extracted
 	stream.setMiddleware(func(src models.Record) (models.Record, error) {
 		recordCount++
+		r.logExtractedRecord(recipe, src)
```
mabdh, line 155:
```suggestion
		r.logger.Info("Successfully extracted record", "record", src.Data().GetResource().Urn, "recipe", recipe.Name)
```

can we do like this directly?

agent/agent.go
```diff
@@ -254,6 +255,7 @@ func (r *Agent) setupSink(ctx context.Context, sr recipe.SinkRecipe, stream *str
 				err = nil
 			}
 		}
+		r.logger.Info("Successfully sank record", "sink", sr.Name, "recipe", recipe.Name)
```
ravisuhag, line 258: sank can be replaced with "published"

StewartJingga (approved):

PR #308, opened 2022-02-02
feat: provide descriptive debugging info on lint

scortier:
### Debugging Information on lint cmd -

Able to detect exact line number : 
- When user provide correct key part in Recipe (without grammar error)

Unable to detect exact line number :
- When key part in recipe is incorrectly passed ( grammar mistake)
- When the user forget to provide required config of plugins.

#### recipe.yaml
```
1. name: kafkaToColumbus
2. source:
3.   name: kafka
4.   config:
5.     brok:
6.     label:
7. sinks:
8.   - name: columbus
9.     config:
10.       host:
11.       test:
12.   - name: xyz
13.   - name: console
```
 #### Output: 
image


cmd/lint.go
```diff
@@ -91,7 +95,102 @@ func LintCmd(lg log.Logger, mt *metrics.StatsdMonitor) *cobra.Command {
...
+		pluginNode := rcp.Source.Node
+		for _, configError := range invalidConfigError.Errors {
+			cfg, ok := pluginNode.Config[configError.Key]
+			if ok {
+				line := cfg.Line
+				fmt.Printf("%s: invalid %s extractor config on line: %d\n", rcp.Name, invalidConfigError.PluginName, line)
+			} else {
+				fmt.Printf("%s: invalid %s extractor config: %s\n", rcp.Name, invalidConfigError.PluginName, configError.Message)
+			}
+		}
```
StewartJingga, line 153: This can be refactored into a function to be used by other plugin types (sink, processor)

recipe/recipe.go
```diff
@@ -1,30 +1,18 @@
...
+	Node       RecipeNode
 }
 
-// ProcessorRecipe contains the json data for a recipe that is being used for
-// generating the processor code for a recipe.
-type ProcessorRecipe struct {
+// PluginRecipe contains the json data for a recipe that is being used for
+// generating the plugins code for a recipe.
+type PluginRecipe struct {
 	Name   string                 `json:"name" yaml:"name" validate:"required"`
```
StewartJingga, line 15: @scortier  This is a breaking changes, old recipes will fail because `source` is still using `type` in the recipe yaml file.

Should we handle backward comp to not break old recipes? We could support both `type` and `name` and then deprecate `type` later.

StewartJingga (changes requested):

recipe/node.go
```diff
@@ -0,0 +1,87 @@
...
+	config := make(map[string]interface{})
+
+	for key, val := range plug.Config {
+		config[key] = val.Value
+	}
+	return config
+}
+
+// toRecipe passes the value from RecipeNode to Recipe
+func (node RecipeNode) toRecipe() (recipe Recipe, err error) {
```
mabdh, line 35: if the operation does not mutate some fields in the struct, we can use reference here instead of value.
```suggestion
func (node *RecipeNode) toRecipe() (recipe Recipe, err error) {
```

utils/config.go
```diff
@@ -1,27 +1,56 @@
...
+			}
+		}
+	}
+	if configErrors == nil {
+		return nil
+	}
+	icErr := plugins.InvalidConfigError{
+		Errors: configErrors,
+	}
+	return icErr
```
mabdh, line 55: We can make the exit point clearer by doing this

```suggestion
	err := validate.Struct(c)
	if err == nil {
	   return nil
	}
	
	var validationErr validator.ValidationErrors
	if errors.As(err, &validationErr) {
	    var configErrors []plugins.ConfigError
	    for _, fieldErr := range validationErr {
	        key := strings.TrimPrefix(fieldErr.Namespace(), "Config.")
		configErrors = append(configErrors, plugins.ConfigError{
			Key:     key,
			Message: fieldErr.Error(),
		})
	    }
	    return plugins.InvalidConfigError{
		Errors: configErrors,
	    }
	}
	return err
```

recipe/node.go
StewartJingga: CMIIW but shouldn't the reason to use pointer is because we want to mutate? 🤔

mabdh: Oh yes that's correct. Should be the other way around. Sorry for mixing this up 😁.
Thanks for pointing out @StewartJingga.
All good here @scortier

recipe/node.go
```diff
@@ -85,3 +86,13 @@ func (node RecipeNode) toSinks() (sinks []PluginRecipe, err error) {
 	}
 	return
 }
+
+// toSupportMultipleSrcTags helps to support both tags `name` and `type` for source
+// till `type` tag gets deprecated
+func (node RecipeNode) toSupportMultipleSrcTags() string {
```
StewartJingga, line 92: this only handles `type` when linting. But what about other commands like `run`? Will source be able to use `name` and `type` interchangeably?

scortier: Yes, its working fine with other cmd too, have checked it.

cmd/lint.go
```diff
@@ -103,81 +105,85 @@ func LintCmd(lg log.Logger, mt *metrics.StatsdMonitor) *cobra.Command {
...
 }
 
-// printPluginError print the plugin type error
-func printPluginError(rcp recipe.Recipe, notFoundError plugins.NotFoundError) {
+// findPluginError print the plugin type error
+func findPluginError(rcp recipe.Recipe, notFoundError plugins.NotFoundError) {
 	if notFoundError.Type == plugins.PluginTypeExtractor {
-		line := rcp.Source.Node.Name.Line
-		fmt.Printf("%s: invalid extractor on line: %d\n", rcp.Name, line)
+		printSourceError(rcp, rcp.Source, notFoundError)
```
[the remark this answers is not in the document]
scortier: So the way backward comp for both tags is supported, it checks which tag is provided by user then prints the Line according to that

StewartJingga (approved):
LGTM

PR #309, opened 2022-02-07
feat(recipe): generate recipes using a template

StewartJingga:


mabdh:
looks good

recipe/generator.go
```diff
@@ -0,0 +1,54 @@
...
+	"path"
+	"text/template"
+)
+
+type FromTemplateData struct {
+	FileName string                 `json:"FileName" yaml:"FileName"`
+	Data     map[string]interface{} `json:"Data" yaml:"Data"`
+}
+
+type FromTemplateConfig struct {
```
mabdh, line 15: not really essential but could the name be just
```suggestion
type TemplateConfig struct {
```
?

Same with the one above as `TemplateData`

mabdh, line 15: not really essential but could the name be just
```suggestion
type TemplateConfig struct {
```
?

Same with the one above as `TemplateData`

mabdh, line 15: not really essential but could the name be just
```suggestion
type TemplateConfig struct {
```
?

Same with the one above as `TemplateData`

mabdh, line 15: not really essential but could the name be just
```suggestion
type TemplateConfig struct {
```
?

Same with the one above as `TemplateData`

recipe/generator.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+	"text/template"
+)
+
+type FromTemplateData struct {
+	FileName string                 `json:"FileName" yaml:"FileName"`
+	Data     map[string]interface{} `json:"Data" yaml:"Data"`
```
bsushmith, line 12: Can we make the tags `Filename` & `Data` to be all small case in the template. would remain consistent with everything else.

recipe/generator.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	err = os.MkdirAll(config.OutputDirPath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("error creating output directory: %w", err)
+	}
+
+	for _, d := range config.Data {
+		if err := outputRecipe(tmplt, config.OutputDirPath, d); err != nil {
+			return fmt.Errorf("error executing template: %w", err)
```
bsushmith, line 35: Can we change this error msg, or return the err as it is - since the line-50 in outputRecipe function is already returning the exact same error.

cmd/gen.go
```diff
@@ -1,80 +1,61 @@
...
+			templatePath := args[0]
+			bytes, err := ioutil.ReadFile(dataFilePath)
+			if err != nil {
+				return err
 			}
 
-			if processors != "" {
-				procList = strings.Split(processors, ",")
+			var data []recipe.FromTemplateData
+			err = yaml.Unmarshal(bytes, &data)
```
scortier, line 44: Can we handle error like this 
```suggestion
			if err = yaml.Unmarshal(bytes, &data); err != nil {
			     return fmt.Errorf("error unmarshalling bytes: %w", err)
			}
```

scortier (changes requested):

recipe/generator.go
StewartJingga: @bsushmith  we could do this, but actually there is a reason why I don't use lowercase for those two fields.

So, to unmarshal a field, it has to be public field, which should be a capital (hence FileName is the struct field name).
We could change the tag to e.g. `yaml: "filename"` so it would be `filename` in the `templatedata.yaml`, but then when you access it from the template you still have to use `FileName` because go template can only read public field from struct.

IMO it could bring more confusion if it was `filename` in `templatedata.yaml` and `FileName` when accessing it in the template.

bsushmith: Makes sense. 👍

bsushmith (approved):

scortier (approved):

mabdh (approved):

PR #311, opened 2022-02-10
feat: handle graceful stream shutdown

scortier:


agent/agent.go
```diff
@@ -156,7 +155,14 @@ func (r *Agent) Run(recipe recipe.Recipe) (run Run) {
 		return src, nil
 	})
 
-	// create a goroutine to let extractor concurrently emit data
+	// a goroutine to shut down stream gracefully
+	go func() {
+		<-ctx.Done()
```
mabdh, line 160: by doing this, are we sure the behaviour would be like this?
```
Close extractor, wait for all queued messaged to reach sinks, and then close all sinks and shutdown meteor.
```

scortier: The approach works in a way that wherever signal is triggered during record extraction till that point , the record count will take place and run cmd info(recipe details) will be shown. I and @StewartJingga have also tested it by creating sample extractor for testing purpose.

wdyt, does the approach seems incorrect or any more detailing we require ?

StewartJingga (approved):

PR #314, opened 2022-02-14
feat: interactive mode in new cmd

ishanarya0:
Users can enable interactive mode by setting the `interactive` (shorthand `i`) flag.

### DIFF
--- a/cmd/new.go
+++ b/cmd/new.go
@@ -118,7 +118,7 @@ func recipeSinkSurvey() ([]string, error) {
 		{
 			Name: "sink",
 			Prompt: &survey.MultiSelect{
-				Message: "What is the sink name?",
+				Message: "Select sink(s)",
 				Options: availableSinks,
 				Help:    "Select the sink(s) for this recipe",
 			},
@@ -147,7 +147,7 @@ func recipeProcessorSurvey() ([]string, error) {
 		{
 			Name: "processor",
 			Prompt: &survey.MultiSelect{
-				Message: "What is the processor name?",
+				Message: "Select processor",
 				Options: availableProcessors,
 				Help:    "Select the processor(s) for this recipe",
 			},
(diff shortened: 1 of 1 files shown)


ravisuhag:
> The current command allows the user to pass blank `source` and `sink` as well, apart from the `processor`. Do we want to allow this? @ravisuhag @StewartJingga

I think, yes, Atleast one source and sink should be selected. @StewartJingga WDYT?

StewartJingga:
> > The current command allows the user to pass blank `source` and `sink` as well, apart from the `processor`. Do we want to allow this? @ravisuhag @StewartJingga
> 
> I think, yes, Atleast one source and sink should be selected. @StewartJingga WDYT?

Yeah a source and a sink at least should be defined. For empty sink I wonder if we can use console as default? will that improve UX? @ravisuhag @ishanarya0

StewartJingga:
Yeah a source and a sink at least should be defined. For empty sink I wonder if we can use console as default? will that improve UX? @ravisuhag @ishanarya0

Not implicitly when running of course, but when generating recipe. So if sink are not mentioned, we set console as the placeholder.

cmd/new.go
```diff
@@ -95,6 +95,7 @@ func NewRecipeCmd() *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&extractor, "extractor", "e", "", "Type of extractor")
+	cmd.MarkFlagRequired("extractor")
```
ravisuhag, line 103: @ishanarya0 Let's disable this rule.

ravisuhag:
@ishanarya0 Instead of users enabling interactive mode. we should ask for the survey if they don't fill in information through flags.

cmd/new.go
```diff
@@ -42,7 +45,8 @@ func NewRecipeCmd() *cobra.Command {
 
 			The recipe will be printed on standard output.
 			Specify recipe name with the first argument without extension.
-			Use commma to separate multiple sinks and processors.`),
+			Use commma to separate multiple sinks and processors.
+			Interactive mode can be enabled by setting --interactive or -i to true`),
```
ravisuhag, line 49: @ishanarya0 we can remove this line now.

PR #318, opened 2022-02-16
feat: fallback to filename if recipe name is missing

ishanarya0:
When using the `gen` command, if `name` is not mentioned in `data`, it will fall back to the `FileName` as the recipe's name.

### DIFF
--- a/recipe/generator_test.go
+++ b/recipe/generator_test.go
@@ -14,41 +14,74 @@ func TestFromTemplate(t *testing.T) {
 	t.Run("should output recipe files using template to output directory", func(t *testing.T) {
 		templatePath := "./testdata/generator/template.yaml"
 		outputDir := "./testdata/generator/temp"
-		data := []recipe.TemplateData{
-			{
-				FileName: "recipe-1",
-				Data: map[string]interface{}{
-					"broker": "main-broker.com:9092",
-					"owner":  "john@example.com",
-				},
-			},
-			{
-				FileName: "recipe-2",
-				Data: map[string]interface{}{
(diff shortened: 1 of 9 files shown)


StewartJingga (changes requested):
The related issue is not only specific to templating but when reading recipe from a file as well.

We need to:
- Fallback to file's name if name is not specified
- Strip the file extension as the recipe name on fallback

Also need help to also write tests for all those changes 🙏 .

ishanarya0:
> The related issue is not only specific to templating but when reading recipe from a file as well.
> 
> We need to:
> 
> * Fallback to file's name if name is not specified
> * Strip the file extension as the recipe name on fallback
> 
> Also need help to also write tests for all those changes 🙏 .

Changes done.

recipe/generator_test.go
```diff
@@ -14,41 +14,74 @@ func TestFromTemplate(t *testing.T) {
...
-				Data: map[string]interface{}{
-					"broker": "main-broker.com:9092",
-					"owner":  "john@example.com",
+
+		t.Run("when recipe has a name", func(t *testing.T) {
+			data := []recipe.TemplateData{
+				{
+					FileName: "recipe-one",
+					Data: map[string]interface{}{
+						"name":   "recipe-1",
```
mabdh, line 23: would it be possible to source these data from the file?
like we do in normal cli?

ishanarya0: Yes, did this. Thanks for the suggestion!

mabdh (approved):

PR #322, opened 2022-02-23
feat(recipe)!: add version support in recipes

GrayFlash:
Deliverables:

- [x] Add version support to recipe reader
- [x] Write tests for reader
- [x] Update previously written templates to version v1beta1
- [x] Update `new` command
- [x] Document changes

### DIFF
--- a/generator/recipe.go
+++ b/generator/recipe.go
@@ -16,6 +16,7 @@ var file embed.FS
 // Template represents the template for generating a recipe.
 type Template struct {
 	Name       string
+	Version    string
 	Source     map[string]string
 	Sinks      map[string]string
 	Processors map[string]string
@@ -26,9 +27,10 @@ var templateFuncs = map[string]interface{}{
 }
 
 // Recipe checks if the recipe is valid and returns a Template
-func Recipe(name string, source string, sinks []string, processors []string) (err error) {
+func Recipe(name string, version string, source string, sinks []string, processors []string) (err error) {
 	tem := Template{
-		Name: name,
+		Name:    name,
+		Version: version,
 	}
 
 	if source != "" {
--- a/generator/recipe.go
+++ b/generator/recipe.go
@@ -26,11 +26,13 @@ var templateFuncs = map[string]interface{}{
 	"indent": indent,
 }
 
+var recipeVersions = [1]string{"v1beta1"}
+
 // Recipe checks if the recipe is valid and returns a Template
-func Recipe(name string, version string, source string, sinks []string, processors []string) (err error) {
+func Recipe(name string, source string, sinks []string, processors []string) (err error) {
 	tem := Template{
 		Name:    name,
-		Version: version,
+		Version: recipeVersions[len(recipeVersions)-1],
 	}
 
 	if source != "" {
@@ -74,4 +76,8 @@ func Recipe(name string, version string, source string, sinks []string, processo
 func indent(spaces int, v string) string {
 	pad := strings.Repeat(" ", spaces)
 	return pad + strings.Replace(v, "\n", "\n"+pad, -1)
+}
+
+func GetRecipeVersions() [1]string {
+	return recipeVersions
 }
--- a/recipe/reader.go
+++ b/recipe/reader.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/odpf/meteor/generator"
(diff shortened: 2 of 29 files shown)


recipe/reader.go
```diff
@@ -94,3 +102,10 @@ func (r *Reader) readDir(path string) (recipes []Recipe, err error) {
 
 	return
 }
+
+func validateRecipeVersion(receivedVersion, expectedVersion string) (err error) {
+	if strings.Compare(receivedVersion, expectedVersion) == 0 {
+		return
+	}
+	return fmt.Errorf("received recipe version %s does not match to expected error %s", receivedVersion, expectedVersion)
```
mabdh, line 110:
```suggestion
	return fmt.Errorf("received recipe version %s does not match to the expected version %s", receivedVersion, expectedVersion)
```
is this a typo?

GrayFlash: Yeah, it was, commiting it.

generator/recipe.go
```diff
@@ -25,10 +26,13 @@ var templateFuncs = map[string]interface{}{
 	"indent": indent,
 }
 
+var recipeVersions = [1]string{"v1beta1"}
```
ishanarya0, line 29: So in this approach, a meteor release will contain an array of recipe versions it can support?
So, in future if we are releasing a meteor that is backward compatible with `v1beta1`, then this array will have multiple versions in it, like `{"v1beta1","v1beta2"}` ?

ravisuhag: Yes.

recipe/reader.go
```diff
@@ -94,3 +102,10 @@ func (r *Reader) readDir(path string) (recipes []Recipe, err error) {
 
 	return
 }
+
+func validateRecipeVersion(receivedVersion, expectedVersion string) (err error) {
+	if strings.Compare(receivedVersion, expectedVersion) == 0 {
+		return
+	}
+	return fmt.Errorf("received recipe version %s does not match to the expected version %s", receivedVersion, expectedVersion)
+}
```
ishanarya0, line 111: So, shall we design this method to match with all the versions in the `recipeVersions` array, instead of just matching with the last element? Then in the future, we will just have to add versions to the `recipeVersions` array.

GrayFlash: So, I think what we need here should be that `v1beta1` will be compatible with `v1beta2` up till `v1` right, and we can throw a warning about the migration to a newer version. And then if there are breaking changes `v2---` comes in and only then we throw an error??

GrayFlash: The idea of matching it with the last element is just to see if it is following the latest recipe version.

And in case there are breaking changes and newer versions, we had a discussion of providing proper alternatives such as migration or version mapping to users.

recipe/reader_test.go
```diff
@@ -221,6 +222,15 @@ func TestReaderRead(t *testing.T) {
 			compareRecipes(t, expected[i], r)
 		}
 	})
+
+	t.Run("should return error if version is missing/incorrect", func(t *testing.T) {
+		reader := recipe.NewReader()
+		_, err := reader.Read("./testdata/missing-version.yaml")
```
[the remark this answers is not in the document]
ravisuhag: @StewartJingga I think the goal is that we want to have all recipes to have this so that we can support upgrades better in future. We can add it but I am thinking since lint will inform it and this is just an additive change and not a change in any configuration. Should it be fine? what do you think? We can mark it breaking in release.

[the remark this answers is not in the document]
StewartJingga: We should definitely mark it as breaking. Will add `!` to the PR name.

Ideally maybe we can start with notice (similar to deprecation notice), but for now I think we can still afford breaking changes as long as we inform all of current meteor users. We are also locking our version so it won't break our deployment.

recipe/reader_test.go
```diff
@@ -1,13 +1,18 @@
...
+	"errors"
 	"os"
 	"testing"
 
 	"github.com/odpf/meteor/recipe"
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	ErrInvalidRecipeVersion = errors.New("recipe version is invalid or not found")
```
StewartJingga, line 13: the test can use the one from recipe package instead of creating it again.

StewartJingga (changes requested):

StewartJingga (approved):
LGTM

PR #323, opened 2022-02-27
feat: pass env variables with --var flag on run cmd

### LINKED ISSUE #228: Support meteor.var or .env for secrets
Instead of meteor_prefix variable accept a .var or .env for accepting secrets. or maybe support both. If not provided, meteor should also promote users for secrets in the command line before running.

GrayFlash:
A pr that resolves #228 

- [x] Allow passing environment variable through a config.yaml file
- [x] Add tests for env secrets
- [x] Update Documentations.


recipe/envs.go
```diff
@@ -1,44 +1,22 @@
...
-
-		key, ok := mapToMeteorKey(key)
-		if !ok {
-			continue
-		}
-
-		data[key] = val
+	data, err := godotenv.Read()
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
GrayFlash: @StewartJingga @mabdh

StewartJingga (changes requested):
1. Need help to also add test cases for the feature 🙏 .
2. Should we try to read from both, env vars and .env file, and if there are conflicts on keys we will choose the more explicit way which is .env file. wdyt? @GrayFlash

GrayFlash:
> 1. Need help to also add test cases for the feature.
Sure, working on it.

> 2. Should we try to read from both, env vars and .env file, and if there are conflicts on keys we will choose the more explicit way which is .env file. wdyt? @GrayFlash

@StewartJingga this makes complete sense. Also, in case of missing .env file I have also added a small warning since it is possible to not have a need of `.env` file at all.

ravisuhag:
@StewartJingga @GrayFlash Do we also want to consider prompting users to pass env in case they are not there in any of the locations(useful for cli usecases)? There can also be a flag that disables prompt and raises an error(useful for CI etc.)

GrayFlash:
> @StewartJingga @GrayFlash Do we also want to consider prompting users to pass env in case they are not there in any of the locations(useful for cli usecases)?

We should pass a prompt or warning for this, incase we don't find it in both .env file and variables.

> There can also be a flag that disables prompt and raises an error(useful for CI etc.)

Do we really want to raise an error? Since a user may want to pass the connection_url maybe for default settings directly to the  recipe, although the warnings should be disabled for CI.

wdyt @ravisuhag @StewartJingga ??

ravisuhag:
> > @StewartJingga @GrayFlash Do we also want to consider prompting users to pass env in case they are not there in any of the locations(useful for cli usecases)? There can also be a flag that disables prompt and raises an error(useful for CI etc.)
> 
> Hmm we could add a prompt, i'm just thinking if specifying variables files would already be enough. So basically by default we would not read variables from a file unless specified. Similar to helm, it only reads if it is specified. e.g.
> 
> ```
> meteor run --var ../variables.yaml
> ```

Make sense. Lets' go ahead with this then.

recipe/envs.go
```diff
@@ -14,8 +14,9 @@ var (
 
 func populateData() map[string]string {
 	dataUpperCase, err := godotenv.Read()
+
 	// warns user about missing .env file
-	if err.Error() == "open .env: no such file or directory" {
+	if err != nil && err.Error() == "open .env: no such file or directory" {
```
ravisuhag, line 19: @GrayFlash we should not rely on string matching and check the type of error.

GrayFlash: Yes agree, will update this one. 

Actually based on a previous comment from @StewartJingga on passing the path to `secrets.yaml`, I have been thinking of adding a flag named `--config` that will allow it.

For e.g.:
`$meteor run _recipes --config secrets.yaml`
where the `--config` flag and path is optional.

And the secrets.yaml if looks like this:

```yaml
MONGO:
  USER_ID: test
  PASS: abcd1234

POSTGRES:
  USER_ID: admin
  PASS: abcd1234
```

And we can read this using the Viper package. wdyt @ravisuhag @StewartJingga ??

ravisuhag: @GrayFlash I am not sure if config is the right flag name for this. Maybe `--var` or `--secrets`. Or any other suggestion?

StewartJingga: yup config does not seem right, we could go `--var` or `-value` to be more explicit.

ravisuhag: +1 on `--var` cc @GrayFlash

GrayFlash: @StewartJingga @ravisuhag have updated the flag name to `--var`

StewartJingga (approved):
LGTM

PR #327, opened 2022-03-09
fix(presto): fix MaxWait and Ping query in test

mabdh:
Fixing flaky test of presto in main branch

### DIFF
--- a/test/utils/dockertest.go
+++ b/test/utils/dockertest.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/ory/dockertest/v3"
 )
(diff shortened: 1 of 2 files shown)


test/utils/dockertest.go
```diff
@@ -13,6 +14,8 @@ import (
 //"retryOp" is an exponential backoff-retry, because the application in the container might not be ready to accept connections yet
 func CreateContainer(opts dockertest.RunOptions, retryOp func(r *dockertest.Resource) error) (purgeFn func() error, err error) {
 	pool, err := dockertest.NewPool("")
+	// exponential backoff-retry, because the application in the container might not be ready to accept connections yet
+	pool.MaxWait = 120 * time.Second
```
StewartJingga, line 18: the default is a minute as seen [here](https://github.com/ory/dockertest/blob/v3/dockertest.go#L547).

Was the `time.Sleep` hack because it could take more than a min for the container to be ready?

mabdh: Yes this is to override the default.

Yes it was a hack since presto db takes some time to startup.

StewartJingga (approved):
LGTM

Issue #326, opened 2022-03-07
Refactor extractors to remove redundancy 


GrayFlash:
## **Problems around redundancy in extractors of current codebase**

* Since we are supporting `conection_url`, there has been a config with just connection_url.
* All extractors follow a similar flow, and hence have many functions operating similarly, e.g.: Info(), init().
* There exist repetitions in the way we write code for similar databases, like MySQL, MSSQL, and Postgres. The only difference is the query being used.

## **Potential Changes that shall help**

Creating a utils package for all extractors at `plugins/extractors/utils`. Using it across all the extractors for functions like Info(), Validate(), New(), and init().

Creating a separate abstraction for similarly working databases, if there are at least 3 or more extractors of similar type. For e.g: clickhouse, MySQL, MSSQL, Postgres can be of a similar category under `sql_based`. As these follow almost the same set of workflow and return columns[] as metadata. These can also be bundled with others like MongoDB, which return columns at least for the buildColumn() function.

PR #328, opened 2022-03-15
refactor: create a util package for extractor

GrayFlash:
Please refer: #326


plugins/extractors/mysql/mysql.go
```diff
@@ -31,31 +31,10 @@ var defaultDBList = []string{
 	"sys",
 }
 
-// Config holds the connection URL for the extractor
-type Config struct {
```
GrayFlash, line 35: So, I intended to remove these structs and move them to utils package.

plugins/extractors/mysql/mysql.go
```diff
@@ -31,31 +31,10 @@ var defaultDBList = []string{
...
 
-// Extractor manages the extraction of data from MySQL
-type Extractor struct {
-	excludedDbs map[string]bool
-	logger      log.Logger
-	config      Config
-	db          *sql.DB
-	emit        plugins.Emit
-}
-
```
GrayFlash, line 49: Same is true for Extractor struct, wanted to move it to `utils/sql_db.go` so that it is not needed to be declared across all the extractors.

But for that I will also need to declare the fields as public, i.e, logger -> Logger.

plugins/extractors/mysql/mysql.go
```diff
@@ -31,31 +31,10 @@ var defaultDBList = []string{
...
-// New returns a pointer to an initialized Extractor Object
-func New(logger log.Logger) *Extractor {
-	return &Extractor{
-		logger: logger,
-	}
-}
-
 // Info returns the brief information about the extractor
-func (e *Extractor) Info() plugins.Info {
```
GrayFlash, line 58: And after these I am confused on how to use receiver (e *Extractor) cause it was throwing error when I tried using (e *extutil.Extractor) since Extractor type is no longer declared in the same package. So I am not sure on whether I should remove it or if there is some other way around.

GrayFlash:
@StewartJingga @mabdh this is pretty much what I was able to move away from `mysql.go` extractor, without hampering any big pattern. I tried moving New() func as well but it keeps throwing errors around logger.Error as well as the return type. So I am not sure how to move it. Kindly review it once, and if it seems fine, I will make similar changes for all the other extractors possible.

plugins/extractors/mysql/mysql.go
```diff
@@ -64,22 +57,17 @@ func (e *Extractor) Info() plugins.Info {
...
 // Init initializes the extractor
 func (e *Extractor) Init(ctx context.Context, configMap map[string]interface{}) (err error) {
-	if err = utils.BuildConfig(configMap, &e.config); err != nil {
+	if err = utils.BuildConfig(configMap, &e.Config); err != nil {
 		return plugins.InvalidConfigError{}
 	}
 
 	// build excluded database list
-	e.buildExcludedDBs()
+	e.ExcludedDbs = extutils.BuildExcludedDBs(defaultDBList)
```
StewartJingga, line 67: ExcludeDBs is not common to all extractors, should let each extractors handle it themselves? wdyt?

plugins/extractors/utils/sql_dbs.go
```diff
@@ -0,0 +1,47 @@
+package utils
```
StewartJingga, line 1: we can rename the filename to `base_extractor.go` instead

plugins/extractors/utils/sql_dbs.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/odpf/meteor/utils"
+)
+
+// Config holds the connection URL for the extractor
+type Config struct {
+	ConnectionURL string `mapstructure:"connection_url" validate:"required"`
+}
+
+// Extractor manages the extraction of data from MySQL
+type BaseExtractor struct {
```
StewartJingga, line 16: This is more like an `BaseSQLExtractor` more than `BaseExtractor`.
An idea is to have `BaseSQLExtractor` it embedded `BaseExtractor`, but I'm still not sure if it is a good idea.
wdyt?

plugins/extractors/mysql/mysql.go
GrayFlash: Do you mean that not every data source has a default list of DB that needs to excluded?

Cause I thought if there are some default dblist for each source, then maybe we can keep the array with their names within each extractor seperately but the BaseExtractor can have it build in util package.

StewartJingga: yes, for example, `kafka`, `bigquery`, `metabase`, `tableau`.

plugins/extractors/utils/sql_dbs.go
GrayFlash: I am not sure if it needs these many Different structs, cause there are not many fields in Extractor struct.

plugins/extractors/mysql/mysql.go
GrayFlash: oh okay, I thought we allowed custom ExcludeDBList from users at one point, which lets users avoid metadata extraction from a particular set of topic/DBs in the source. But if there are no things at this point then I'll move it back to extractors.

plugins/extractors/utils/sql_dbs.go
ravisuhag: I am thinking should it be a struct or can we actually just have pure util functions?

mabdh: Hi @GrayFlash, I don't think that we need that much generalization/abstraction by pulling this into a struct called `BaseExtractor`. Our `Plugin` interface and `Extractor` interface are abstract enough to define the contract.

It is true those functions are repetitive (the one in BaseExtractor). We could just have some pure and generic function to do that functions. 

For example, rather than having a function called `BuildExcludedDBs`, maybe we can generalize it to `BuildExclusions` / `BuildStringExclusion`.

We could improve the util package naming by not naming it with generic name like `util` but more into with the name that indicate the purpose of functions inside it.

GrayFlash:
I have added utils that fetch a list of databases, a list of tables and help build excluded DB map. Have also used it for `Postgres` and `MySql`. Refactoring further where we obtain metadata and build it as per the proto is tough since each metadata extractor has a different combination of fields that we are extracting and it is not much of help to add a util function where we just run query and do not process the result.

If things seem good, I can use the utils functions in other extractors as well `presto`, `oracle`, `mssql`, `mariadb`, and `clickhouse`.

plugins/extractors/postgres/postgres.go
```diff
@@ -25,12 +26,11 @@ import (
 //go:embed README.md
 var summary string
 
-var defaultExcludes = []string{"information_schema", "root", "postgres"}
+var defaultDBList = []string{"information_schema", "root", "postgres"}
 
 // Config holds the set of configuration options for the extractor
 type Config struct {
 	ConnectionURL string `mapstructure:"connection_url" validate:"required"`
-	Exclude       string `mapstructure:"exclude"`
```
GrayFlash, line 33: I will restore this one, but wanted to know if it something we want to allow with every other extractor as well.?? Or if there is some reason to have it only for Postgres. 

cc @StewartJingga @scortier

plugins/utils/sql_utils.go
```diff
@@ -0,0 +1,58 @@
+package utils
```
StewartJingga, line 1: I would rename the package to be `sqlutil` (notice the singular as well).
wdyt? @GrayFlash @mabdh @ravisuhag 

so using it will be like e.g `sqlutil.FetchDBs`.
also reason for `sqlutil` instead of `sqlutils` is to follow go base library convention to use singular whenever possible.

StewartJingga: I know package name can be aliased when importing, but I think we can even do it from the package itself instead of aliasing it everywhere.

mabdh: Yes that would be more readable

GrayFlash: Yes @StewartJingga makes sense

StewartJingga (approved):
LGTM

PR #331, opened 2022-03-20
feat: allow agent level config with --config flags

GrayFlash:


cmd/run.go
```diff
@@ -56,6 +57,13 @@ func RunCmd(lg log.Logger, mt *metrics.StatsdMonitor, cfg config.Config) *cobra.
 			"group:core": "true",
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if configFile != "" {
+				var err error
+				cfg, err = config.LoadFromPath(configFile)
+				if err != nil {
+					return err
```
mabdh, line 64: why not just

```suggestion
				cfg, err := config.LoadFromPath(configFile)
				if err != nil {
					return err
```
?

config/config.go
```diff
@@ -29,3 +29,15 @@ func Load() (cfg Config, err error) {
```
mabdh, line 29: I think we could also give a config sample `meteor.yaml.sample` in the repo

cmd/run.go
GrayFlash: Wouldn’t that make `cfg` a local variable as well??

mabdh:
@GrayFlash perhaps we could also have a new command in Makefile to `copy-config`

```
copy-config:
   cp ./config/meteor.yaml.sample ./meteor.yaml
```

config/meteor.yaml.sample
```diff
@@ -0,0 +1,7 @@
+LOG_LEVEL: hehehehehe
```
mabdh, line 1: we can make the default to `info` or `debug` 😄 
```suggestion
LOG_LEVEL: debug
```

config/meteor.yaml.sample
```diff
@@ -0,0 +1,7 @@
...
+STATSD_ENABLED: false
+STATSD_HOST: "localhost:8005"
+STATSD_PREFIX: lol
```
mabdh, line 4:
```suggestion
STATSD_PREFIX: meteor
```

config/meteor.yaml.sample
```diff
@@ -0,0 +1,7 @@
+LOG_LEVEL: hehehehehe
+STATSD_ENABLED: false
+STATSD_HOST: "localhost:8005"
```
mabdh, line 3:
```suggestion
STATSD_HOST: "localhost:8125"
```

8125 is statsd default port

config/meteor.yaml.sample
GrayFlash: sorry, i wasn’t meant to push it, it was just to test the working 😅😅

config/config.go
```diff
@@ -18,10 +18,17 @@ type Config struct {
 	StopOnSinkError             bool   `mapstructure:"STOP_ON_SINK_ERROR" default:"false"`
 }
 
-func Load() (cfg Config, err error) {
-	err = config.
-		NewLoader(config.WithPath("./")).
-		Load(&cfg)
+func Load(configFile string) (cfg Config, err error) {
+	if configFile != "" {
```
ravisuhag, line 22: @GrayFlash If flag has a default already this check will not be needed.

mabdh (approved):

PR #333, opened 2022-03-22
fix: broken plugin tests on ci

### LINKED ISSUE #329: Some extractor plugin tests are breaking and not executed in the workflow CI
**Describe the bug**
There were some build tags `integration` for each plugin extractor tests but we renamed some of the tags to `plugins`. In the workflow, we only run `plugin_test` in the main branch. Even with this, not all of them are tagged with `plugins`, this caused some plugins are not being covered to test in the `plugin_test` workflow.
Some plugins have broken test
- Superset: port health check is `8080`, but the exposed docker port is `8088`
- Tableau: Ownership is populated (due to columbus sink changes)
- There might be other plugins test that will be broken too if we update the tag to `plugins`

**To Reproduce**
1. Update all go build tags in `plugins/extractor` to `plugins`
2. Run the `make test-plugins`
3. Test plugin will throw errors

**Expected behavior**
No error when running `make test-plugins` after updating all tags to `plugins`

GrayFlash:
Closes #329


GrayFlash:
Have enabled Plugin Test on the current branch for testing plugins on CI, need to **delete** it before PR is merged.

GrayFlash:
Plugins that are currently failing in ci and need to be fixed:

- [ ] bigtable
- [ ] clickhouse
- [ ] kafka
- [ ] superset

mabdh:
`bigtable`, `clickhouse`, `kafka` are still broken. I wonder if it is caused by the misconfigured `dockertest`

mabdh (approved):

PR #337, opened 2022-04-01
feat(columbus): update columbus api to new patch api

scortier:


plugins/sinks/columbus/columbus.go
```diff
@@ -1,20 +1,26 @@
 package columbus
 
 type Record struct {
```
StewartJingga, line 3: we can rename this to be more clear e.g. `RequestPayload`. wdyt?

plugins/sinks/columbus/sink_test.go
```diff
@@ -39,14 +38,10 @@ func TestInit(t *testing.T) {
...
 			},
-			{
-				"host": host,
-				"type": "",
-			},
 		}
 		for i, config := range invalidConfigs {
 			t.Run(fmt.Sprintf("test invalid config #%d", i+1), func(t *testing.T) {
-				columbusSink := columbus.New(newMockHTTPClient(config, http.MethodGet, url, []columbus.Record{}), testUtils.Logger)
+				columbusSink := columbus.New(newMockHTTPClient(config, http.MethodGet, url, columbus.Record{}), testUtils.Logger)
```
StewartJingga, line 44: it should be `PATCH` instead of `GET`

plugins/sinks/columbus/sink_test.go
```diff
@@ -313,33 +339,43 @@ func TestSink(t *testing.T) {
...
 			},
 		},
 	}
 
 	for _, tc := range successTestCases {
 		t.Run(tc.description, func(t *testing.T) {
-			tc.expected.Data = tc.data
-			payload := []columbus.Record{tc.expected}
+			tc.expected.Asset.Data = tc.data
+			fmt.Println(tc.expected)
```
StewartJingga, line 371: dont forget to remove this

StewartJingga (changes requested):

plugins/sinks/columbus/columbus.go
scortier: `Record` specify more about what actually model is about as it is known to be a payload only, while `RequestPayload` is a general term for payload request for any api . So should we change it ?

StewartJingga: yes that worked before, but now Columbus does not have Record and only has Asset.

StewartJingga (approved):
LGTM

PR #339, opened 2022-04-12
feat(postgres): extract table access for users

### LINKED ISSUE #335: feat(postgres): extract user access
I need to have more metadata from postgres such as db users that have access to the table/dataset.

Users access information on table metadata using postgres extractor

GrayFlash:
closes #335

### DIFF
--- a/plugins/extractors/postgres/postgres_test.go
+++ b/plugins/extractors/postgres/postgres_test.go
@@ -1,3 +1,6 @@
+//go:build plugins
+// +build plugins
+
 package postgres_test
 
 import (
--- a/plugins/extractors/postgres/postgres_test.go
+++ b/plugins/extractors/postgres/postgres_test.go
@@ -10,8 +10,12 @@ import (
 	"os"
 	"testing"
 
+	"github.com/odpf/meteor/models"
+	commonv1beta1 "github.com/odpf/meteor/models/odpf/assets/common/v1beta1"
+	facetsv1beta1 "github.com/odpf/meteor/models/odpf/assets/facets/v1beta1"
 	assetsv1beta1 "github.com/odpf/meteor/models/odpf/assets/v1beta1"
 	"github.com/odpf/meteor/test/utils"
+	ut "github.com/odpf/meteor/utils"
 
 	"database/sql"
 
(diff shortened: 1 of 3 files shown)


plugins/extractors/postgres/postgres_test.go
```diff
@@ -145,3 +143,76 @@ func execute(db *sql.DB, queries []string) (err error) {
...
+						Length:     20,
+					},
+				},
+			},
+			Properties: &facetsv1beta1.Properties{
+				Attributes: ut.TryParseMapToProto(map[string]interface{}{
+					"grants": []interface{}{
+						map[string]interface{}{
+							"user":            "test_user",
+							"privilege_types": "INSERT,SELECT,UPDATE,DELETE,TRUNCATE,REFERENCES,TRIGGER",
```
GrayFlash, line 211: @StewartJingga I tried to use `[]string{"SELECT", "INSERT"}` here, but haven't been able to since the `utils.TryParseMapToProto()` function throws error.

If []string not supported by proto yet??

StewartJingga: @GrayFlash  btw you should do it like this instead
```
"grants": []map[string]interface{}{
  {
    "user": "test_user",
    ...,
  },
}
```

GrayFlash: Not sure but when I use it this way, I was getting this error:

`panic: proto: invalid type: []map[string]interface {}`

GrayFlash: Also I went through this comment on protobuf, I'm checking if there has been any further addition to data types.

https://github.com/golang/protobuf/issues/1302#issuecomment-805453221

plugins/extractors/postgres/postgres_test.go
```diff
@@ -174,7 +171,7 @@ func getExpected() []models.Record {
 					"grants": []interface{}{
 						map[string]interface{}{
 							"user":            "test_user",
-							"privilege_types": "INSERT,SELECT,UPDATE,DELETE,TRUNCATE,REFERENCES,TRIGGER",
+							"privilege_types": []interface{}{"INSERT", "SELECT", "UPDATE", "DELETE", "TRUNCATE", "REFERENCES", "TRIGGER"},
```
GrayFlash, line 177: @StewartJingga thought of this to support an array of privileges a user has. wdyt???

StewartJingga (changes requested):

utils/custom_properties.go
```diff
@@ -65,6 +65,14 @@ func appendCustomFields(metadata models.Metadata, customFields map[string]interf
 	return properties, err
 }
 
+func ConvertStringListToInterface(s []string) []interface{} {
```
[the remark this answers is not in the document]
GrayFlash: I thought it might be helpful incase we later need to use MapToProto in other places, so kept it generic like abduh suggested last time we were refactoring for the sqlutils package.

But since here we are using it only once, moving it to Postgres.

PR #347, opened 2022-05-24
fix: misleading error for missing version field

GrayFlash:


recipe/reader.go
```diff
@@ -99,7 +99,7 @@ func (r *Reader) readDir(path string) (recipes []Recipe, err error) {
 	for _, entry := range entries {
 		recipe, err := r.readFile(filepath.Join(path, entry.Name()))
 		if err != nil {
-			continue
```
GrayFlash, line 102: @StewartJingga the cause of this issue seems to be this `continue` statement when reading files from a directory. What happens is while reading a list of recipes we were skipping the recipes which had an issue or error in them, and continued running the rest.

But when there is just one file in the directory, and that too is invalid, it just reports the error as recipe not found.

StewartJingga: Nice, catch!

Also, we can try wrapping the most the errors there using `fmt.Errorf` to enable debugging easier.

recipe/reader.go
```diff
@@ -99,7 +99,7 @@ func (r *Reader) readDir(path string) (recipes []Recipe, err error) {
...
+			lg.Error(err.Error())
```
StewartJingga, line 102: We can do it like this
```
lg.Warn("skipping file", "path", path, "err", err.Error())
```

recipe/reader.go
StewartJingga: btw we still need the continue syntax if we don't want it to break on invalid recipe.

If you want to break, we can just return error instead of logging it

recipe/reader.go
```diff
@@ -31,14 +31,14 @@ func NewReader(pathToConfig string) *Reader {
 }
 
 //  Read loads the list of recipes from a give file or directory path.
-func (r *Reader) Read(path string) (recipes []Recipe, err error) {
+func (r *Reader) Read(lg log.Logger, path string) (recipes []Recipe, err error) {
```
StewartJingga, line 34: instead of passing logger on read, we can pass it when initiating reader via `NewReader`

StewartJingga (changes requested):

recipe/reader.go
GrayFlash: We are not breaking it even now, instead of `continue` we are logging the error, and it continues reading other recipes in the directory.

StewartJingga: CMIIW if we are not continuing, it will continue to next code, which will put empty recipe on recipe list.

GrayFlash: Ah Yes, makes sense.

StewartJingga (approved):
LGTM

PR #354, opened 2022-06-01
feat: add Instance_label to extractors

### LINKED ISSUE #353: feat: add `instance_label` to extractors
Similar to https://github.com/odpf/meteor/issues/352, do the same for the following extractors:
- postgres
- mysql
- mssql
- tableau

Below extractors have `instance_label` config:
- postgres
- mysql
- mssql
- tableau

GrayFlash:
closes #353 

- [x] postgres
- [x] mysql
- [x] mssql
- [x] tableau


mabdh:
> @StewartJingga I have made changes to add `instance_label` as a part of config for the above 4 extractors. Wanted to know if `sessionID` instead of password&username is to be added for `tableau`. I am not sure if sql extractors will also need a sessionID since we use connection string there.

It is called `auth token` in tableau and not `session id`. If metabase has it, nothing is stopping us to add the similar flow for tableau.

GrayFlash:
@mabdh @StewartJingga created a separate issue for it here. #358

StewartJingga (approved):

ravisuhag:
@GrayFlash @StewartJingga Do you think we can find a better name for this config instead of `instance_label` ?

GrayFlash:
can we use `record_label` since we use it to identify records being extracted as a part of `urn`?? wdyt @ravisuhag @StewartJingga

mabdh:
What about `namespace`?

GrayFlash:
@mabdh @StewartJingga @ravisuhag  even `namespace` seems fine in terms of conveying its purpose.

StewartJingga:
> @mabdh @StewartJingga @ravisuhag even `namespace` seems fine in terms of conveying its purpose.

`namespace` works, but I feel like it has a chance of having another meaning on some source. (e.g. `tableau`).

mabdh:
`namespace` naming is actually inspired by this [URN IETF doc](https://datatracker.ietf.org/doc/html/rfc8141#section-2.2). `Namespace Specific String (NSS)`

GrayFlash:
@mabdh makes sense.

@StewartJingga @ravisuhag should we use `namespace` then??

ravisuhag:
We can use `identifier` as well to address the concern @StewartJingga  has. cc @mabdh @GrayFlash

PR #357, opened 2022-06-06
refactor(bigquery): add table name to filter logadmin query

mabdh:
- add table name to filter logadmin query
- collect log per table instead of building a global hash map

### DIFF
--- a/plugins/extractors/bigquery/bigquery.go
+++ b/plugins/extractors/bigquery/bigquery.go
@@ -103,13 +103,15 @@ func (e *Extractor) Init(ctx context.Context, configMap map[string]interface{}) 
 	}
 
 	if e.config.IsCollectTableUsage {
-		errL := e.galClient.Init(ctx, auditlog.Config{
-			ProjectID:           e.config.ProjectID,
-			ServiceAccountJSON:  e.config.ServiceAccountJSON,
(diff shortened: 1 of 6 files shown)


mabdh:
Need to increase coverage to `bigquery/auditlog`

plugins/extractors/bigquery/bigquery.go
```diff
@@ -181,10 +174,20 @@ func (e *Extractor) extractTable(ctx context.Context, ds *bigquery.Dataset, emit
 
 // Build the bigquery table metadata
 func (e *Extractor) buildTable(ctx context.Context, t *bigquery.Table, md *bigquery.TableMetadata) *assetsv1beta1.Table {
+	var tableStats *auditlog.TableStats
+	if e.config.IsCollectTableUsage {
+		// Fetch and extract logs first to build a map
+		var errL error
+		tableStats, errL = e.galClient.Collect(ctx, t.TableID)
```
StewartJingga, line 181: CMIIW, so now we will fetch table usage once for each table, which will create more requests but we won't fetch unused data/rows? @mabdh

mabdh: @StewartJingga yes correct

StewartJingga (approved):
Code's LGTM

PR #360, opened 2022-06-07
fix(elastic): fix failing tests due to unsorted indexes

### LINKED ISSUE #359: fix(elasticsearch): inconsistent order of extracted records
**Describe the bug**
The indices of elasticsearch are sometimes not in ascending order, also the columns. Thus leading to failing tests on our CI

**To Reproduce**
run `make test-plugins` locally for a few times, and the test will break.

**Expected behavior**
The test should pass consistently without false-negatives.

GrayFlash:
closes #359

### DIFF
--- a/plugins/extractors/elastic/elastic.go
+++ b/plugins/extractors/elastic/elastic.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"sort"
 
 	"github.com/pkg/errors"
 
@@ -118,7 +117,6 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error) 
 		for i := range docProperties {
 			columNames = append(columNames, i)
 		}
-		sort.Strings(columNames)
 		for _, i := range columNames {
 			columns = append(columns, &facetsv1beta1.Column{
 				Name:     i,
--- a/plugins/extractors/elastic/elastic.go
+++ b/plugins/extractors/elastic/elastic.go
@@ -105,7 +105,6 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error) 
 	var indexes []string
 	for x.Next() {
 		indexName := x.Key().String()
-		// fmt.Println(indexName)
 		indexes = append(indexes, indexName)
 	}
 	sort.Strings(indexes)
--- a/plugins/extractors/elastic/elastic.go
+++ b/plugins/extractors/elastic/elastic.go
@@ -101,30 +101,22 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error) 
 		return
 	}
 	x := reflect.ValueOf(r["indices"]).MapRange()
-	var indexes []string
 	for x.Next() {
 		indexName := x.Key().String()
-		indexes = append(indexes, indexName)
-	}
-	for _, indexName := range indexes {
-		docProperties, err1 := e.listIndexInfo(indexName)
+		docProperties, err1 := e.listIndexInfo(x.Key().String())
 		if err1 != nil {
 			err = err1
 			return
 		}
 		var columns []*facetsv1beta1.Column
-		var columNames []string
 		for i := range docProperties {
-			columNames = append(columNames, i)
-		}
-		for _, i := range columNames {
 			columns = append(columns, &facetsv1beta1.Column{
 				Name:     i,
 				DataType: docProperties[i].(map[string]interface{})["type"].(string),
 			})
 		}
 		countRes, err1 := e.client.Search(
-			e.client.Search.WithIndex(indexName),
+			e.client.Search.WithIndex(x.Key().String()),
 		)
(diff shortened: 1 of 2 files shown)


plugins/extractors/elastic/elastic.go
```diff
@@ -101,22 +102,32 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error)
 		return
 	}
 	x := reflect.ValueOf(r["indices"]).MapRange()
+	var indexes []string
 	for x.Next() {
 		indexName := x.Key().String()
-		docProperties, err1 := e.listIndexInfo(x.Key().String())
+		indexes = append(indexes, indexName)
+	}
+	sort.Strings(indexes)
```
StewartJingga, line 110: does the indices order also cause the flaky test, or is it just columns order?

plugins/extractors/elastic/elastic.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"sort"
```
mabdh, line 9: @GrayFlash if it is to solve the test problem. why don't we just sort the expected and the actual data in test?

plugins/extractors/elastic/elastic.go
```diff
@@ -101,22 +102,32 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error)
...
 	}
 	x := reflect.ValueOf(r["indices"]).MapRange()
+	var indexes []string
 	for x.Next() {
 		indexName := x.Key().String()
-		docProperties, err1 := e.listIndexInfo(x.Key().String())
+		indexes = append(indexes, indexName)
+	}
+	sort.Strings(indexes)
+	for _, index := range indexes {
```
StewartJingga, line 111: let's rename `index` to `indexName`, `index` can mean another thing in a for loop.

StewartJingga (changes requested):

plugins/extractors/elastic/elastic.go
StewartJingga: that actually works too 🤔

plugins/extractors/elastic/elastic.go
GrayFlash: yes there were some instance where it was out of order

plugins/extractors/elastic/elastic.go
GrayFlash: Thought we should maintain the consistency in order of records being emmited. Usually all other extractor's client follow a certain pattern in showing the result of query.

mabdh: Hmm but if the problem is in the test case (flaky test) and there is no requirements to sort the extracted records, adding this in the main code would just add some additional unnecessary computation. 🤔

GrayFlash: @mabdh was able to fix it for indexes actually in the test part, but I can seem to find a function to get the data in `Schema` from records. There doesn’t seem to be any function in metadata model interface for it. So it is still being handled by sorting the columns before emitting. Should work?? Or else we might need some changes in proto models itself.

GrayFlash: please let me know if I'm missing something here. 😅

mabdh: In that case maybe we can add `GetSchema` here? @StewartJingga 
https://github.com/odpf/meteor/blob/5f7cd1471bc68ede73b4e1206856373d84cf654e/models/metadata.go#L11

GrayFlash: hey, folks any further comment on this? shall I move on with what I suggested?

mabdh: @GrayFlash IMO, if the purpose of the test is not in the `order of columns` or `number of extracted columns should be equal to the number provided columns`. Then it does fine to reduce the dummy data to 1.

GrayFlash: @StewartJingga @mabdh for now I have removed one column from dummy data and tested it for the same, if we end up adding `GetSchema`, we can fix it other way.

plugins/extractors/elastic/elastic.go
```diff
@@ -101,22 +101,30 @@ func (e *Extractor) Extract(ctx context.Context, emit plugins.Emit) (err error)
 		return
 	}
 	x := reflect.ValueOf(r["indices"]).MapRange()
+	var indexes []string
```
StewartJingga, line 104: do we need this `indexes`? we can directly process the index in the `for x.Next()` block right?

StewartJingga (changes requested):

plugins/extractors/elastic/elastic.go
GrayFlash: yes, we can do that. removed all the changes from elastic.go btw, made them just to sort the index and columns. but they are not necessary anymore.

StewartJingga (approved):

PR #363, opened 2022-06-13
feat(optimus): expose configuring grpc max receive size

StewartJingga:
new `max_size_in_mb1` in `optimus` extractor to set max megabytes for GRPC client to receive message.
```
source:
  name: optimus
  config:
    host: optimus.com
    max_size_in_mb: 100
```

### DIFF
--- a/plugins/extractors/optimus/optimus_test.go
+++ b/plugins/extractors/optimus/optimus_test.go
@@ -38,7 +38,7 @@ func TestInit(t *testing.T) {
 		ctx := context.TODO()
 
 		client := new(mockClient)
-		client.On("Connect", ctx, validConfig["host"]).Return(nil)
+		client.On("Connect", ctx, validConfig["host"], 0).Return(nil)
 		defer client.AssertExpectations(t)
 
 		extr := optimus.New(testutils.Logger, client)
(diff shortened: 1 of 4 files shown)


mabdh (approved):

plugins/extractors/optimus/optimus_test.go
```diff
@@ -118,7 +118,7 @@ func (c *mockClient) GetJobTask(
 }
 
 func setupExtractExpectation(ctx context.Context, client *mockClient) {
-	client.On("Connect", ctx, validConfig["host"]).Return(nil).Once()
+	client.On("Connect", ctx, validConfig["host"], 0).Return(nil).Once()
```
GrayFlash, line 121: have we tested it for any value other tha 0 mb??

GrayFlash:
Rest LGTM 🚀

plugins/extractors/optimus/optimus_test.go
StewartJingga: we can put other value, but right now there is no way of test the impact unless we have lots of testdata. @GrayFlash

GrayFlash: okay, seems good for now then

PR #368, opened 2022-06-13
feat: add stencil sink

scortier:
Currently supporting schema format :

- [x] JSON
- [x] AVRO


plugins/sinks/stencil/sink.go
```diff
@@ -0,0 +1,223 @@
...
+		Type:       JsonType(resource.GetType()),
+		Properties: properties,
+	}
+
+	return record, nil
+}
+
+func (s *Sink) buildJsonProperties(table *assetsv1beta1.Table) (properties []map[string]Property) {
+	columns := table.GetSchema().GetColumns()
+	if columns == nil {
```
StewartJingga, line 160: let's do `len(columns) > 0` instead. Nil slice will return 0 for length.

plugins/sinks/stencil/sink.go
```diff
@@ -0,0 +1,223 @@
...
+			dataType = JsonTypeArray
+		case "object":
+			dataType = JsonTypeObject
+		case "bool", "boolean":
+			dataType = JsonTypeBoolean
+		default:
+			dataType = JsonTypeString
+		}
+
+		columnRecord[column.Profile.String()] = Property{
```
StewartJingga, line 183: we will only need `columnRecord[column.Name]`, we don't need the others.

plugins/sinks/stencil/sink.go
```diff
@@ -0,0 +1,223 @@
...
+		Schema:     "https://json-schema.org/draft/2020-12/schema",
+		Title:      resource.GetName(),
+		Type:       JsonType(resource.GetType()),
+		Properties: properties,
+	}
+
+	return record, nil
+}
+
+func (s *Sink) buildJsonProperties(table *assetsv1beta1.Table) (properties []map[string]Property) {
```
StewartJingga, line 158: it should only return `map[string]Property`.

StewartJingga (changes requested):

plugins/sinks/stencil/sink_test.go
```diff
@@ -0,0 +1,346 @@
...
+						Description: "It is the user ID",
+					},
+					"description": {
+						Type:        []stencil.JsonType{stencil.JsonTypeString, stencil.JsonTypeNull},
+						Description: "It is the description",
+					},
+				},
+			},
+		},
+		//{
```
ravisuhag, line 169: @scortier If this piece of code is not needed, let's remove this instead of commenting.

scortier: Sure, will add more test and refactor remaining part, working on it

PR #370, opened 2022-06-16
fix: sink all records to file after stream closes

### LINKED ISSUE #366: file sink overwrites rather than append
**Describe the bug**
File sink overwrites the file rather than appending records to it. 

Example: `optimus` extractor emits record for every job, and if `file` sink plugin is used. At the end of recipe run, the file will contain only one job record(the last one), rather than all the jobs. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a sample file sink recipe. eg: using optimus as extractor and file as sink
2. run the recipe
3. check the records in the file.

**Expected behaviour**
The file should contain all records in the `file` sink.

GrayFlash:
closes #366


agent/agent.go
```diff
@@ -179,6 +182,11 @@ func (r *Agent) Run(ctx context.Context, recipe recipe.Recipe) (run Run) {
 
 	// start listening.
 	// this process is blocking
+	defer func() {
```
spy16, line 185: this still seems incorrect. from the place where the sink is setup (assuming line 146 where `r.setupSink()` is being called) to this line, there are multiple `return` in-between. what happens when we end-up returning early?

correct way to do this is usually to have the `defer Close()` at the site of initialisation of this resource (after error checking)

```golang
sink, err := r.setupSink(...)
if err != nil {
   return err
}
defer sink.Close()
```

GrayFlash: @spy16 thanks for this, will take note for future usecase as well

mabdh:
@GrayFlash are we making `append` vs `overwrite` configurable by user?

plugins/sinks/file/file.go
```diff
@@ -65,10 +66,13 @@ func (s *Sink) Init(ctx context.Context, config map[string]interface{}) (err err
 }
 
 func (s *Sink) Sink(ctx context.Context, batch []models.Record) (err error) {
-	var data []models.Metadata
```
mabdh, line 68: What if, instead of having a global variable, we do collect as it is and write in this Sink (not in the Close part).

We can write in this step by Opening an `append-only` or `write-only` file like [this](https://yourbasic.org/golang/append-to-file/).

GrayFlash: hey, @mabdh actually the issue here is with the output format being `json` we cannot append data to the file directly, hence we discussed keeping it in a way that replaces the file instead of making a config for `append`. Thus we maintain a variable to store the data and save it to the file only once at `Close()`.

mabdh: For json, actually we can if we use http://ndjson.org/ like format

GrayFlash: Yes, sure thing I will look more into it actually. @mabdh

GrayFlash: @mabdh have updated with `ndjson` format and seems to work fine to me. Working to make the file append/replace configurable for user now. Let me know if there are any concerns

plugins/sinks/file/file.go
```diff
@@ -60,7 +62,16 @@ func (s *Sink) Init(ctx context.Context, config map[string]interface{}) (err err
 	if err := s.validateFilePath(s.config.Path); err != nil {
 		return err
 	}
+
 	s.format = s.config.Format
+	if s.config.Overwrite {
+		s.File, err = os.Create(s.config.Path)
+		return err
+	}
+	s.File, err = os.OpenFile(s.config.Path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0777)
```
mabdh, line 71: with this i think we still need to close the file? are we closing the opened file somewhere?

GrayFlash: yeah, missed it earlier, closing the IO to file in the sink.Close() function.

plugins/sinks/file/file.go
```diff
@@ -60,7 +62,16 @@ func (s *Sink) Init(ctx context.Context, config map[string]interface{}) (err err
 	if err := s.validateFilePath(s.config.Path); err != nil {
```
mabdh, line 62: do we still need to do this? inferring format from the extension.

it looks like our `format` config is not working cause of this.

GrayFlash: hey, more context on what is not working??
I basically use this function to ensure that filename is valid, like it has extension. I am not inferring the format from this function.

mabdh: I am trying with this config
```
      path: postgres_food_app_data.xxx
      format: ndjson
```
but the output is in yaml

plugins/sinks/file/file.go
```diff
@@ -69,8 +80,8 @@ func (s *Sink) Sink(ctx context.Context, batch []models.Record) (err error) {
 	for _, record := range batch {
 		data = append(data, record.Data())
 	}
-	if s.format == "json" {
-		err := s.jsonOut(data)
+	if s.format == "ndjson" {
```
mabdh, line 83: we need to update the sample config too if we update this format

StewartJingga:
When there is an `existing` file on `overwrite` mode, it will now remove the `existing` file on initialization.

What it means is, even before sinking process, if there is an issue on other plugins configuration, we would already remove the `existing` file.

@bsushmith will that behaviour be expected?

if not, one thing we could do it, we can sink the result to a temporary file, and once everything is done, we replace the `existing` file with the `temp` file.

ravisuhag:
> 

@StewartJingga I think that should be fine as the user has already taken the decision of overriding it. Extraction is happening post that decision.

plugins/sinks/file/file.go
GrayFlash: hey mas, can you check again, It is working for me.

mabdh (approved):

PR #372, opened 2022-06-17
feat: add metric collection to catch sink errors

### LINKED ISSUE #365: feat: improve metric gathering to catch sink error
Meteor now only collects metrics on recipe level. The issue is when there is an error on sink, the error is only being logged and no metric is being gathered. This can be misleading because recipe can be marked as success but with lots of error on sink part.

Catch sink error metrics so we can setup alert and better monitoring.

GrayFlash:
closes #365


agent/run.go
```diff
@@ -14,11 +14,18 @@ const (
 	TaskTypeSink TaskType = "sink"
 )
 
+type RunSink struct {
+	Error   error               `json:"error"`
+	Success bool                `json:"success"`
```
mabdh, line 19: do we really need both `success` and `error` here? can we just have either one?
if `error` nil, it must be success or we can just use success only.

metrics/statsd.go
```diff
@@ -45,6 +45,28 @@ func (m *StatsdMonitor) RecordRun(run agent.Run) {
...
+		fmt.Sprintf(
+			"%s.%s,name=%s,success=%s,source=%s,sink=%s",
+			m.prefix,
+			"sinkStatus",
+			run.Recipe.Name,
+			successText,
+			run.Recipe.Source.Name,
+			sink.Recipe.Name,
+		),
+	)
```
mabdh, line 69: can use this directly
```suggestion
	m.client.Increment(
		fmt.Sprintf(
			"%s.%s,name=%s,success=%t,source=%s,sink=%s",
			m.prefix,
			"sinkStatus",
			run.Recipe.Name,
			sink.Success,
			run.Recipe.Source.Name,
			sink.Recipe.Name,
		),
	)
```

metrics/statsd.go
```diff
@@ -45,6 +45,28 @@ func (m *StatsdMonitor) RecordRun(run agent.Run) {
...
+func (m *StatsdMonitor) RecordSink(run agent.Run, sink agent.RunSink) {
+	var successText = "false"
+	if sink.Success {
+		successText = "true"
+	}
+	m.client.Increment(
+		fmt.Sprintf(
+			"%s.%s,name=%s,success=%s,source=%s,sink=%s",
+			m.prefix,
+			"sinkStatus",
```
mabdh, line 63: we can make this as const to be consistent like other metrics. see the definiton on top of this file.

agent/run.go
mabdh: I see. Even for that I think it is redundant to define success and error altogether. We can just use error and avoid ambiguity for that.

But it is fine I guess for now.

metrics/statsd.go
```diff
@@ -12,10 +12,11 @@ import (
 	"github.com/odpf/meteor/recipe"
 )
 
-var (
+const (
 	runDurationMetricName    = "runDuration"
 	runRecordCountMetricName = "runRecordCount"
 	runMetricName            = "run"
+	sinkMetricName           = "sinkStatus"
```
StewartJingga, line 19: let's generalise it to plugin and keep same format like run recipe
```suggestion
	pluginRunMetricName           = "runPlugin"
```

metrics/statsd.go
```diff
@@ -47,6 +48,21 @@ func (m *StatsdMonitor) RecordRun(run agent.Run) {
 	)
 }
 
+// RecordSink records a individual sinks behavior in a run
+func (m *StatsdMonitor) RecordSink(recipeName, sourceName string, sink agent.PluginRun) {
```
StewartJingga, line 52: let's generalise this to `RecordPlugin`, and let's use parameters only.

We only need, `recipe name`, `plugin name`, `success`, `plugin type`.

We can skip source name because we already have recipe name, that should be enough.

agent/run.go
```diff
@@ -14,6 +14,12 @@ const (
 	TaskTypeSink TaskType = "sink"
 )
 
+type PluginRun struct {
```
StewartJingga, line 17: let's not create a struct here if it is only for metric gathering purposes only.

StewartJingga (changes requested):

mabdh (approved):

ravisuhag:
@GrayFlash Can you please resolve merge conflicts o this one?

PR #385, opened 2022-07-11
fix: broken build worflow

GrayFlash:
Tasks:
- [x] fix broken build CI
- [x] update homebre-tap for brew

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - main
-      - fix-build-worflow
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - main
+      - fix-build-worflow
```
GrayFlash, line 8: not to be merged unless this is removed.

GrayFlash:
Upgrading all our workflows to golang `1.17` seems to fix it for now.
Build on this branch was successful: https://github.com/odpf/meteor/runs/7276758660?check_suite_focus=true

PR #396, opened 2022-08-16
fix: skip config load for cmds that don't require it

### LINKED ISSUE #394: bug: config not found message on all commands that do not require it
**Describe the bug**
I keep getting below message when running all meteor commands that does not require config file
```
unable to find config file, loading from env and defaults: open ./meteor.yaml: no such file or directory
```

**To Reproduce**
1. Make sure you don't have `meteor.yaml`
2. Run `meteor list extractors`
3. See message

**Expected behavior**
Should not show `unable to find config file` message for below commands that do not require it:
1. meteor list
2. meteor info
3. meteor gen
4. others...

sudo-suhas:
- Load config only for 'lint' and 'run' commands. This would ensure that
  there are no warnings logs related to config file not being found for
  commands that do not require it.
- In case the config file is not present, print the error but proceed
  with defaults.
- Remove metrics/instrumentation for lint command since it does not
  provide any value.

Closes #394

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,42 +1,13 @@
 package cmd
 
 import (
-	"fmt"
-	"os"
-
 	"github.com/MakeNowJust/heredoc"
-	"github.com/odpf/meteor/config"
-	"github.com/odpf/meteor/metrics"
-	"github.com/odpf/meteor/plugins"
 	"github.com/odpf/salt/cmdx"
-	"github.com/odpf/salt/log"
 	"github.com/spf13/cobra"
 )
 
-const exitError = 1
-
 // New adds all child commands to the root command and sets flags appropriately.
 func New() *cobra.Command {
-	cfg, err := config.Load("./meteor.yaml")
-	if err != nil {
-		fmt.Printf("ERROR: %s\n", err.Error())
-		os.Exit(1)
-	}
-
-	lg := log.NewLogrus(log.LogrusWithLevel(cfg.LogLevel))
-	plugins.SetLog(lg)
-
-	// Setup statsd monitor to collect monitoring metrics
-	var mt *metrics.StatsdMonitor
-	if cfg.StatsdEnabled {
-		client, err := metrics.NewStatsdClient(cfg.StatsdHost)
-		if err != nil {
-			fmt.Printf("ERROR: %s\n", err.Error())
-			os.Exit(exitError)
-		}
-		mt = metrics.NewStatsdMonitor(client, cfg.StatsdPrefix)
-	}
-
 	var cmd = &cobra.Command{
 		Use:           "meteor <command> <subcommand> [flags]",
 		Short:         "Metadata CLI",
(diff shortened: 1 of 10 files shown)


sudo-suhas:
```
BREAKING CHANGE: Config file not found errors were being ignored
earlier for 'meteor run' but would fail now if the config file is not
found.
```

Not 100% sure if this behaviour would be correct. More than happy to roll this back if we want to allow `meteor run` without any config file being found. However, if the config file is not found, even the default values won't be applied:

```go
package main

import (
	"fmt"

	"github.com/odpf/meteor/config"
)

func main() {
	cfg, err := config.Load("./doesnotexist.yaml")
	fmt.Printf("cfg: %#v\nerr: %s\n", cfg, err)
}
```

```

➜ go run ./.playground/cfgload.go                                                                                         
cfg: config.Config{LogLevel:"", StatsdEnabled:false, StatsdHost:"", StatsdPrefix:"", MaxRetries:0, RetryInitialIntervalSeconds:0, StopOnSinkError:false}
err: unable to find config file, loading from env and defaults: open ./doesnotexist.yaml: no such file or directory
```

ravisuhag:
@sudo-suhas Checkout this PR as well. https://github.com/odpf/shield/pull/155 We should follow a similar approach on this. cc @mabdh @StewartJingga

cmd/root.go
```diff
@@ -18,23 +19,23 @@ const exitError = 1
...
-		client, err := metrics.NewStatsdClient(cfg.StatsdHost)
-		if err != nil {
-			fmt.Printf("ERROR: %s\n", err.Error())
-			os.Exit(exitError)
-		}
-		mt = metrics.NewStatsdMonitor(client, cfg.StatsdPrefix)
+	mt, err := newStatsdMonitor(cfg)
+	if err != nil {
+		fmt.Printf("ERROR: %s\n", err.Error())
+		os.Exit(exitError)
```
mabdh, line 38: Rather than throwing `os.Exit(exitError)` here in `cmd/root`, I think it is better to handler `os.Exit` in `main.go` file. So maybe this New function could also return error and this line could just simply `return err`. wdyt? @sudo-suhas @StewartJingga

sudo-suhas:
> @sudo-suhas Checkout this PR as well. https://github.com/odpf/shield/pull/155 We should follow a similar approach on this.

@ravisuhag could you elaborate on this a bit? Are you suggesting that we should use `github.com/odpf/salt/cmdx` for loading the config?

cmd/root.go
sudo-suhas: Although this specific instance is no longer present, have done the suggested change for possible error during config load.

ravisuhag:
> > @sudo-suhas Checkout this PR as well. [odpf/shield#155](https://github.com/odpf/shield/pull/155) We should follow a similar approach on this.
> 
> @ravisuhag could you elaborate on this a bit? Are you suggesting that we should use `github.com/odpf/salt/cmdx` for loading the config?

I think that as well, and also handling of missing config approach.

sudo-suhas:
@ravisuhag @StewartJingga have done the changes as per our discussion and also updated the PR desc.

ravisuhag (approved):

mabdh (approved):

sudo-suhas:
In the sync review/QA for changes, @StewartJingga recommended to still print error for `run` and `lint` commands if config file is not present. Have done this change.

PR #405, opened 2022-08-30
fix: accept and use scope for generating recipe

### LINKED ISSUE #402: bug: meteor "new recipe" command is not using latest recipe requirement
**Describe the bug**
Current `meteor new` command does not use the latest recipe requirements.

generated-recipe.yaml
```
name: sample
version: v1beta1
source:
  type: kafka
  config:     
    broker: "my-broker:6668"
sinks:
  - name: http
    config:       
      # The url (hostname and route) of the http service
      url: http://localhost:3333
      method: POST
      # Additional HTTP headers, multiple headers value are separated by a comma
      headers:
        X-Other-Header: value1, value2
```

running this get below error
```
ERRO[0000] error running recipe                          duration_ms=0 err="failed to setup extractor: could not initiate extractor \"kafka\": urn scope is required to generate unique urn" recipe=sample records_count=0
ERRO[0000] failed to setup extractor: could not initiate extractor "kafka": urn scope is required to generate unique urn
```

The error is because since version `0.5.0`, meteor introduce new required `source.scope` field on recipe.
This field is needed to make sure asset's `urn` is unique by using `scope` as some sort of namespacing feature.

**To Reproduce**
1. Run `meteor new recipe sample -e kafka -s http > recipe.yaml`
2. Copy printed out recipe to a file (`recipe.yaml`)
3. Run `meteor run recipe.yaml`

**Expected behavior**

### Update `new` command
`meteor new recipe` should include `scope` on the generated recipe. It could also make `scope` as part of required input when running the command. e.g. `--scope=my-scope`.

### Docs updated
Somewhere in the docs, it would be helpful if the new `scope` field is being mentioned so that users are not stuck.

sudo-suhas:
Add new mandatory flag --scope to 'meteor new recipe' command for
creating a sample recipe and use the provided scope for generating the
recipe.

BREAKING CHANGE: Method signature for generator.Recipe() changed.

Closes #402.

### DIFF
--- a/cmd/new.go
+++ b/cmd/new.go
@@ -30,6 +30,7 @@ func NewCmd() *cobra.Command {
 func NewRecipeCmd() *cobra.Command {
 	var (
 		extractor  string
+		scope      string
 		sinks      string
 		processors string
 	)
@@ -88,13 +89,24 @@ func NewRecipeCmd() *cobra.Command {
 				}
 			}
 
-			return generator.Recipe(args[0], extractor, sinkList, procList)
+			return generator.Recipe(generator.RecipeParams{
+				Name:       args[0],
+				Source:     extractor,
+				Scope:      scope,
+				Sinks:      sinkList,
+				Processors: procList,
+			})
 		},
 	}
 
 	cmd.Flags().StringVarP(&extractor, "extractor", "e", "", "Type of extractor")
+	cmd.Flags().StringVarP(&scope, "scope", "n", "", "URN's namespace")
 	cmd.Flags().StringVarP(&sinks, "sinks", "s", "", "List of sink types")
 	cmd.Flags().StringVarP(&processors, "processors", "p", "", "List of processor types")
+
+	if err := cmd.MarkFlagRequired("scope"); err != nil {
+		panic(err)
+	}
 
 	return cmd
 
--- a/generator/recipe.go
+++ b/generator/recipe.go
@@ -13,11 +13,15 @@ import (
 //go:embed recipe.yaml
 var file embed.FS
 
-// Template represents the template for generating a recipe.
-type Template struct {
-	Name       string
-	Version    string
-	Source     map[string]string
+// templateData represents the template for generating a recipe.
+type templateData struct {
+	Name    string
+	Version string
+	Source  struct {
+		Name         string
+		Scope        string
+		SampleConfig string
+	}
 	Sinks      map[string]string
 	Processors map[string]string
 }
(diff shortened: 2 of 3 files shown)


generator/recipe.go
```diff
@@ -15,9 +15,13 @@ var file embed.FS
 
 // Template represents the template for generating a recipe.
 type Template struct {
```
sudo-suhas, line 17: any concern with making this struct private? technically change to this struct is also breaking change..

generator/recipe.go
```diff
@@ -15,9 +15,13 @@ var file embed.FS
...
-	Name       string
-	Version    string
-	Source     map[string]string
+	Name    string
+	Version string
+	Source  struct {
+		Type         string
```
StewartJingga, line 21: We can rename `Type` to `Name` since we want to move away from `Type`. wdyt?

cmd/new.go
```diff
@@ -88,14 +89,25 @@ func NewRecipeCmd() *cobra.Command {
...
+				Source:     extractor,
+				Scope:      scope,
+				Sinks:      sinkList,
+				Processors: procList,
+			})
 		},
 	}
 
 	cmd.Flags().StringVarP(&extractor, "extractor", "e", "", "Type of extractor")
+	cmd.Flags().StringVarP(&scope, "scope", "n", "", "URN namespace")
```
StewartJingga, line 103: `scope` might not always be URN at the moment.

StewartJingga (changes requested):

generator/recipe.go
StewartJingga: no concern atm, since it is only being used inside the package, we can make it as private.

cmd/new.go
sudo-suhas: could you suggest an appropriate description?

generator/recipe.go
sudo-suhas: sure, will do

cmd/new.go
StewartJingga: Oh actually I misunderstood what you mean by `URN namespace`, my bad.

We can use `URN's namespace` or keeping it as it is also works for me. wdyt?

sudo-suhas: cool, works

StewartJingga (approved):

generator/recipe.yaml
```diff
@@ -1,9 +1,10 @@
 name: {{.Name}}
 version: {{.Version}}
 source:
-{{- range $key, $value := .Source }}
-  type: {{$key}}
-  config: {{$value | indent 4}}
+{{- with .Source }}
+  type: {{.Name}}
```
StewartJingga, line 5: oh here is as well, we can change `type` to `name`. 🙏

StewartJingga (changes requested):

generator/recipe.yaml
sudo-suhas: Done, apologies for missing it earlier.

StewartJingga (approved):

PR #406, opened 2022-09-01
fix(gen): format, fix sample configs used for generating recipes

sudo-suhas:
- Introduce and use template helper function on sample configs of
  sinks, extractors and processors to replace tabs, if present, with
  spaces in sample configs. If tabs are present, parsing the file as
  YAML fails.
- Fix sample configs for sinks/http, extractors/{kafka, tableau}.
- Reformat sample configs for consistency.

### DIFF
--- a/generator/recipe.go
+++ b/generator/recipe.go
@@ -27,7 +27,8 @@ type templateData struct {
 }
 
 var templateFuncs = map[string]interface{}{
-	"indent": indent,
+	"indent":          indent,
+	"fmtSampleConfig": fmtSampleConfig,
 }
 
 var recipeVersions = [1]string{"v1beta1"}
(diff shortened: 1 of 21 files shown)


sudo-suhas:
@StewartJingga this issue is present in #408 as well. Can we review and merge this change?

generator/recipe.go
```diff
@@ -93,6 +94,14 @@ func indent(spaces int, v string) string {
 	return pad + strings.Replace(v, "\n", "\n"+pad, -1)
 }
 
+func fmtSampleConfig(s string) string {
```
ravisuhag, line 97: @sudo-suhas lets rename this to something short? maybe `cfmt` ?

sudo-suhas: Does rawfmt work?

StewartJingga:
Should we move sample config to its own file 🤔 ? or even better ideally we should be able to generate sample config from the `Config` struct itself. wdyt @ravisuhag @sudo-suhas

StewartJingga:
> Should we move sample config to its own file 🤔 ? or even better, maybe we should generate sample config from the `Config` struct itself. wdyt @ravisuhag @sudo-suhas

We can keep current change for now, need more analysis and effort for generating from `Config` struct but I think it is possible. Just something to think about.

StewartJingga (approved):

generator/recipe.go
sudo-suhas: addressed in https://github.com/odpf/meteor/pull/412, please check

PR #408, opened 2022-09-06
feat: add shield sink

ishanarya0:


plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,195 @@
...
+	url := fmt.Sprintf("%s/admin/v1beta1/users/%s", s.config.Host, record.Email)
+	req, err := http.NewRequest(http.MethodPut, url, bytes.NewBuffer(payloadBytes))
+	if err != nil {
+		return
+	}
+
+	for hdrKey, hdrVal := range s.config.Headers {
+		hdrVals := strings.Split(hdrVal, ",")
+		for _, val := range hdrVals {
+			req.Header.Add(hdrKey, val)
```
sudo-suhas, line 112: Should we trim the spaces from `val`?

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,195 @@
...
+
+		s.logger.Info("successfully sinked record to shield", "record", metadata.Name)
+	}
+
+	return
+}
+
+func (s *Sink) Close() (err error) { return }
+
+func (s *Sink) send(record RequestPayload) (err error) {
```
sudo-suhas, line 96: Is there any reason we are not using context while creating the HTTP request?

sudo-suhas (changes requested):

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,195 @@
...
+	payloadBytes, err := json.Marshal(record)
```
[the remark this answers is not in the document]
ishanarya0: I think it will, tested it in go playground [here](https://go.dev/play/p/sj4J64vyqHs). WDYT?

Yes, I think it is not a good practice to use named params in case of a few params, so we can avoid it.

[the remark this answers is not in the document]
sudo-suhas: will leave it up to you but personally, I found it a little confusing and it is not really clear to me what value is being provided by use of named return parameter.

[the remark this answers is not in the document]
ishanarya0: Yes, that makes sense. Let's get rid of it.

plugins/sinks/shield/sink.go
ishanarya0: Yes, we should.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return
+	}
+
+	// send request
+	url := fmt.Sprintf("%s/admin/v1beta1/users/%s", s.config.Host, record.Email)
```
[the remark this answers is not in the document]
ishanarya0: Sure, it looks good!

plugins/sinks/shield/sink.go
ishanarya0: @StewartJingga I saw this missing in other places as well, any particular reason for this?

StewartJingga: no particular reason actually, could've been missed before. Let's pass context whenever's possible here.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,198 @@
...
+	Description: "Send user information to shield http service",
+	Summary:     summary,
+	Tags:        []string{"http", "sink"},
+	SampleConfig: heredoc.Doc(`
+	# The hostname of the shield service
+	host: https://shield.com
+	# Additional HTTP headers send to shield, multiple headers value are separated by a comma
+	headers:
+	  X-Shield-Email: meteor@odpf.io
+      X-Other-Header: value1, value2
```
sudo-suhas, line 43:
```suggestion
	  X-Other-Header: value1, value2
```

`heredoc` will strip the tab at the start, so we can keep the sample config neatly formatted.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,198 @@
...
+func (s *Sink) Close() (err error) { return }
+
+func (s *Sink) send(ctx context.Context, record RequestPayload) error {
+	payloadBytes, err := json.Marshal(record)
+	if err != nil {
+		return err
+	}
+
+	// send request
+	url := fmt.Sprintf("%s/admin/v1beta1/users/%s", url.PathEscape(s.config.Host), url.PathEscape(record.Email))
```
sudo-suhas, line 104: Don't think we need to apply `PathEscape` to the hostname. I don't think it would be possible to have a hostname that has special characters - https://stackoverflow.com/questions/3523028/valid-characters-of-a-hostname.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,198 @@
...
+			val = strings.TrimSpace(val)
+			req.Header.Add(hdrKey, val)
+		}
+	}
+
+	res, err := s.client.Do(req)
+	if err != nil {
+		return err
+	}
+	if res.StatusCode == 200 {
```
sudo-suhas, line 122: Can we close the response here? Context - https://stackoverflow.com/questions/18598780/is-resp-body-close-necessary-if-we-dont-read-anything-from-the-body/18601625#18601625

You could use the following helper:

```go
// DrainBody drains and closes the response body to avoid the following
// gotcha:
// http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body
func DrainBody(resp *http.Response) {
	if resp == nil {
		return
	}

	_, _ = io.Copy(ioutil.Discard, resp.Body)
	_ = resp.Body.Close()
}
```

Or as suggested by @spy16 in the discussion thread on slack:

```go
type withCleanup struct{ io.ReadCloser }

func (irc *withCleanup) Close() error {
	if irc.ReadCloser == nil {
		return nil
	}
	_, _ = io.Copy(ioutil.Discard, irc.ReadCloser)
	err := irc.ReadCloser.Close()
	irc.ReadCloser = nil
	return err
}
```

> then right after i get response from client, i do `resp.Body = &withCleanup{ resp.Body }` (so normal `defer resp.Body.Close()` ends up doing the proper cleanup sequence)

Article on this gotcha and more: https://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,198 @@
...
+	data := resource.GetData()
+
+	mapdata, err := s.buildShieldData(data)
+	if err != nil {
+		return RequestPayload{}, err
+	}
+
+	name, ok := mapdata["full_name"].(string)
+	if !ok {
+		return RequestPayload{}, errors.New("name must be a string")
```
sudo-suhas, line 151: We could improve the error here by including the information about unexpected type with `%T`.

sudo-suhas (changes requested):
_Apologies for missing some of the comments in the previous review_

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,198 @@
...
+	`),
+}
+
+type httpClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+type Sink struct {
+	plugins.BasePlugin
+	client httpClient
```
StewartJingga, line 53: We can use GRPC to connect to shield. Here is an example in optimus extractor.
https://github.com/odpf/meteor/blob/main/plugins/extractors/optimus/optimus.go

StewartJingga (changes requested):

plugins/sinks/shield/sink.go
spy16: +1

plugins/sinks/shield/sink.go
spy16: if we have a generic client wrapper, then i would suggest using `withCleanup` approach since the `resp.Body.Close()` would be happening somewhere else..

but if we are directly using the client, then we `defer DrainBody(resp)` would be the way to go 👍

plugins/sinks/shield/sink.go
ishanarya0: Iinfo seems to be working as expected.

image

Can you elaborate on your concern?

sudo-suhas: suggesting to keep the raw text neatly formatted.

plugins/sinks/shield/sink.go
ishanarya0: Changed to GRPC-based client.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,207 @@
...
+		s.logger.Info("successfully sinked record to shield", "record", metadata.Name)
+	}
+
+	return nil
+}
+
+func (s *Sink) Close() (err error) {
+	return
+	//TODO: Connection closes even when some records are unpiblished
+	//TODO: return s.client.Close()
```
StewartJingga, line 99: you can call `s.client.Close()` in this hook

StewartJingga (changes requested):
Need help updating this docs as well 🙏 
https://github.com/odpf/meteor/blob/main/docs/docs/reference/sinks.md#sinks

plugins/sinks/shield/sink.go
ishanarya0: Can't do this, as this leads to the early closing of the connection.
This is also mentioned [here](https://github.com/odpf/meteor/blob/9de6002d1abd7c8ca2ee0f631a4d6ab8c46014a6/agent/agent.go#L283)

ravisuhag:
@ishanarya0 CI is failing because of code coverage. Can you check this?

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,207 @@
...
+	if err := s.client.Connect(ctx, s.config.Host); err != nil {
+		return fmt.Errorf("error connecting to host: %w", err)
+	}
+
+	return nil
+}
+
+func (s *Sink) Sink(ctx context.Context, batch []models.Record) error {
+	for _, record := range batch {
+		metadata := record.Data()
```
StewartJingga, line 79:
```suggestion
		asset := record.Data()
```

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,207 @@
...
+			return err
+		}
+	} else {
+		err = fmt.Errorf("not able to parse error returned %v", err)
+	}
+
+	return err
+}
+
+func (s *Sink) buildShieldPayload(resource *assetsv1beta2.Asset) (RequestPayload, error) {
```
StewartJingga, line 150:
```suggestion
func (s *Sink) buildShieldPayload(asset *assetsv1beta2.Asset) (RequestPayload, error) {
```

StewartJingga: also we don't need `RequestPayload` anymore if we are using gRPC.
we can directly use `UpdateUserRequest` struct from Shield proto.

So instead of `Asset` -> `RequestPayload` -> `UpdateUserRequest`,
we can do `Asset` -> `UpdateUserRequest`.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,207 @@
...
+	data := resource.GetData()
+
+	mapdata, err := s.buildShieldData(data)
```
StewartJingga, line 153: we can rename `buildShieldData` to `getUserData` which might return error if `anypb.Any` is not a `User` struct.

And if it returns error that Asset is not a User, we can just skip processing it since `shield` only concerns about `User` data.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,207 @@
...
+		Name:     name,
+		Email:    email,
+		Metadata: metadata,
+	}
+
+	return record, nil
+}
+
+func (s *Sink) buildShieldData(anyData *anypb.Any) (map[string]interface{}, error) {
+	var mapData map[string]interface{}
```
StewartJingga, line 183: Instead of `mapData` we can directly unmarshal it to `v1beta2.User`

```
import "github.com/odpf/meteor/models/odpf/assets/v1beta2"

var user v1beta2.User
err := anyData.UnmarshalTo(&user)
```

StewartJingga (changes requested):

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (s *Sink) Sink(ctx context.Context, batch []models.Record) error {
+	for _, record := range batch {
+		asset := record.Data()
+		s.logger.Info("sinking record to shield", "record", asset.GetUrn())
+
+		userRequestBody, err := s.buildUserRequestBody(asset)
+		if err != nil {
+			return errors.Wrap(err, "failed to build shield payload")
```
StewartJingga, line 82: we can just `continue` here if it has error.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (s *Sink) buildUserRequestBody(asset *assetsv1beta2.Asset) (*sh.UserRequestBody, error) {
+	data := asset.GetData()
+
+	var user assetsv1beta2.User
+	err := data.UnmarshalTo(&user)
+	if err != nil {
+		return &sh.UserRequestBody{}, errors.Wrap(err, "not a User struct")
+	}
```
StewartJingga, line 141: Can we move it to another method? basically if this line fails, we can just skip the record instead of failing the sink or the whole batch.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,180 @@
...
+		return &sh.UserRequestBody{}, errors.New("empty user name")
+	}
+	if user.Email == "" {
+		return &sh.UserRequestBody{}, errors.New("empty user email")
+	}
+	if user.Attributes == nil {
+		return &sh.UserRequestBody{}, errors.New("empty user attributes")
+	}
+
+	metadataBytes, err := json.Marshal(user.Attributes)
```
StewartJingga, line 153: if im not mistaken `user.Attributes` is already structpb.Struct.

we can directly pass it to `UserRequestBody.Metadata` without needing to marshal/unmarshal it.

StewartJingga (changes requested):

plugins/sinks/shield/sink.go
StewartJingga: Basically the goal is, if `Asset.Data` is not User then skip the whole process, instead of error.

ishanarya0: Understood, let me put a 'continue` in the loop of the Sink method, [line 82](https://github.com/odpf/meteor/blob/a97ad7997936c512ad18e20f073bc654d2ffd01a/plugins/sinks/shield/sink.go#L82) and [line 86](https://github.com/odpf/meteor/blob/a97ad7997936c512ad18e20f073bc654d2ffd01a/plugins/sinks/shield/sink.go#L86). This shall skip the users which have failure building the request or sending.

ishanarya0: Not on line 86, as we still need to send retry errors in case of send failure.

plugins/sinks/shield/sink.go
ishanarya0: Oh right, my bad. Fixing this.

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (s *Sink) Sink(ctx context.Context, batch []models.Record) error {
+	failedRequestBodyBuildCount := 0
+	for _, record := range batch {
+		asset := record.Data()
+		s.logger.Info("sinking record to shield", "record", asset.GetUrn())
+
+		userRequestBody, err := s.buildUserRequestBody(asset)
+		if err != nil {
+			s.logger.Error(errors.Wrap(err, "failed to build shield payload").Error(), "record", asset.Name)
```
StewartJingga, line 81:
```suggestion
			s.logger.Error("failed to build shield payload", "err", err, "record", asset.Name)
```

plugins/sinks/shield/sink.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		if err = s.send(ctx, userRequestBody); err != nil {
+			return errors.Wrap(err, "error sending data")
+		}
+
+		s.logger.Info("successfully sinked record to shield", "record", asset.Name)
+	}
+
+	if failedRequestBodyBuildCount > 0 {
```
StewartJingga, line 94: for now we can remove this

StewartJingga (changes requested):

StewartJingga (approved):

PR #411, opened 2022-09-12
fix(bigquery): panic-ing due to invalid type for structpb

StewartJingga:


plugins/extractors/bigquery/bigquery.go
```diff
@@ -150,10 +150,9 @@ func (e *Extractor) extractTable(ctx context.Context, ds *bigquery.Dataset, emit
 	tb := ds.Tables(ctx)
 	for {
 		table, err := tb.Next()
-		if err == iterator.Done {
+		if errors.Is(err, iterator.Done) || errors.Is(err, context.Canceled) {
 			break
-		}
-		if err != nil {
+		} else if err != nil {
```
sudo-suhas, line 155: `else` is not required here.

agent/agent.go
```diff
@@ -171,7 +171,7 @@ func (r *Agent) Run(ctx context.Context, recipe recipe.Recipe) (run Run) {
 	go func() {
 		defer func() {
 			if r := recover(); r != nil {
-				run.Error = fmt.Errorf("%s", r)
+				run.Error = fmt.Errorf("recover run panic: %s", r)
```
sudo-suhas, line 174: could we change the error message to something more meaningful? something like `agent run: close stream: panic: %s`

sudo-suhas (changes requested):

plugins/extractors/bigquery/bigquery.go
StewartJingga: it would take one extra line right if we remove `else`?

sudo-suhas: Sometimes things are clearer with more lines of code. In this case, my intention was to keep it similar to the recommendation to avoid an else condition when we return from inside the if condition.

that said, don't have a strong opinion though, okay with it either way.

StewartJingga: valid concern, but in this exact case I think it is still fine.

will keep it like this for now to resolve this 🙏 .

Issue #413, opened 2022-09-20
Add HTTP Extractor


ishanarya0:
Build a generic HTTP extractor.

A sample config 

```
source:
  name: http
  config:
    request:
      host: http://example.com
      method: POST
      path: /v1/employees
      encoding: json
      query:
        foo: bar
      body: {
        "fieldA": "valueA",
        "nestedField": {
          "fieldB": "valueB"
        }
      }
    response:
      model: user
      mapping: // explore with jsonpath
        urn: 
        name: "name"
        age: "profile.age"
        array:
          - id: "array[].id"
            name: "array[].name"
        data:
          attributes:
            foo: "nested.foo"
      expected_codes:
        - "200"
        - "201"
      retry_codes:
        - "500"
        - "503"
```

PR #414, opened 2022-09-20
feat: add http extractor

ishanarya0:
Building a generic HTTP extractor #413


ishanarya0:
We decided to re-analyze and implement this, so closing this one. 
cc: @sudo-suhas @StewartJingga

PR #415, opened 2022-09-23
fix: send asset metadata create_time & update_time to Compass

### LINKED ISSUE #409: bug(compass): create_time and update_time are not being sent
Our current `Asset` model has below fields to store metadata create & update time.

```go
// The timestamp when the object was created.
CreateTime *timestamppb.Timestamp `json:"create_time,omitempty"`

// The timestamp when the object was last modified.
UpdateTime *timestamppb.Timestamp `json:"update_time,omitempty"`
```
Which will be populated by extractor when's possible.

But when sinking to `compass`, these `CreateTime` and `UpdateTime` are not being sent.

This is because `Compass` request payload does not have any fields to accomodate this except `Data` field which is basically just a `map[string]interface{}`. And when sending to `Compass`, Meteor sets `Data` field directly from `Asset.Data` which is a concrete model that does not have those time fields as well. 

### 1. Move time fields from Asset to Table, Dashboard, Topic, etc.

This will modify Meteor's `Table | Dashboard | Topic | ...` models to have `create_time` and `update_time` fields while removing it from `Asset`.

When sinking to `compass`, these `create_time` and `update_time` will be sent along inside `Data` field in `compass` payload.

### 2. Open proposal in odpf/compass to have those time fields in its Asset model.

If `compass.Asset` model has these fields available, then Meteor would just need to pass those time fields directly from its `Asset`.

```go
compassAsset := Asset{
  ...other fields
  URN: meteorAsset.GetUrn(),
  CreateTime: meteorAsset.CreateTime,
  UpdateTime: meteorAsset.UpdateTime,
}
```

sudo-suhas:
Compass request data structure did not support sending the create_time and update_time parsed from data present in extractors. For example, this would be the time a table was created in a database and the time it was last updated at.

- Update Proton's commit ID to pull in changes that add create_time and update_time fields in each of the asset metadata types - table, user, topic etc.
- Populate Asset.Data.{CreateTime, UpdateTime} and stop populating Asset.{CreateTime,UpdateTime} (these fields are used by the system to track the first and last injestion run that inserted/updated the asset).

Depends on https://github.com/odpf/proton/pull/191.
Closes #409.


sudo-suhas:
@StewartJingga this PR can be reviewed now. Have verified changes to the best of my ability. Will remove the draft status once https://github.com/odpf/proton/pull/191 is merged and the proton commit is updated.

StewartJingga (approved):

PR #427, opened 2022-10-29
feat: add application_yaml extractor

sudo-suhas:
- Add application_yaml extractor for reading the given YAML file and emitting an Asset of type service.
- Fix relative link to contributing guide in plugin READMEs.
- Remove unused code in utils/config.go
- Move private helper functions in caramlstore extractor to plugin pkg for
  - parsing Kafka scope from bootstrap servers
  - constructing URN from BigQuery table's fully qualified name.

[DONE] ~~_Needs https://github.com/odpf/proton/pull/214 to be merged first (update `PROTON_COMMIT` in the Makefile after that)_~~


plugins/extractors/service_yaml/service.go
```diff
@@ -0,0 +1,84 @@
...
+	Inputs      []Dataset         `json:"inputs" yaml:"inputs" validate:"dive"`
+	Outputs     []Dataset         `json:"outputs" yaml:"outputs" validate:"dive"`
+	CreateTime  time.Time         `json:"create_time" yaml:"create_time"`
+	UpdateTime  time.Time         `json:"update_time" yaml:"update_time"`
+	Labels      map[string]string `json:"labels" yaml:"labels"`
+}
+
+type Dataset struct {
+	Type       string                 `json:"type" yaml:"type" validate:"required,oneof=stream batch"`
+	Platform   string                 `json:"platform" yaml:"platform" validate:"required,oneof=kafka bigquery"`
```
StewartJingga, line 30: should we limit this field to only kafka and bigquery? or is it better if we leave it open?
`kafka` `bigquery` seems to be our setup only.

Also if we are talking about input, CMIIW it can come from other services as well. (worker hitting service's APIs)

plugins/extractors/service_yaml/README.md
```diff
@@ -0,0 +1,109 @@
...
+      topic: "estimate-log" # REQUIRED
+      bootstrap_servers: "host:port" # REQUIRED
+outputs: # OPTIONAL
+  - type: "stream"
+    platform: "kafka"
+    options:
+      topic: "string"
+      bootstrap_servers: "host:port"
+create_time: "2006-01-02T15:04:05Z07:00"
+update_time: "2006-01-02T15:04:05Z07:00"
```
StewartJingga, line 52: These sample dates is unable to be parsed, we can remove the time offset

```suggestion
create_time: "2006-01-02T15:04:05Z"
update_time: "2006-01-02T15:04:05Z"
```

plugins/extractors/service_yaml/service.go
ravisuhag: Also, let's call this extractor `application`.

plugins/extractors/service_yaml/build_asset.go
```diff
@@ -0,0 +1,130 @@
...
+
+	downstreams, err = buildLineageResources(svc.Outputs)
+	if err != nil {
+		return nil, nil, fmt.Errorf("build lineage downstreams: %w", err)
+	}
+
+	return upstreams, downstreams, nil
+}
+
+func buildLineageResources(dd []Dataset) ([]*v1beta2.Resource, error) {
```
ravisuhag, line 75: @sudo-suhas should this be kept outside of the extractor? I assume this can go as part of the processor? I think this is a very specific usecase.

plugins/extractors/service_yaml/service.go
sudo-suhas: > or is it better if we leave it open?

How should we handle building lineage in that case? It might be surprising if only a specific subset is supported for building lineage. We could expand the list of supported platforms in this PR as well if you have something specific in mind.

> Also, let's call this extractor `application`.

@ravisuhag could you elaborate a bit on why? An `application` extractor that reads a file to emit a `Service` asset doesn't seem intuitive. Or are you suggesting we rename `{service => application}` in all places?

One more thing I would like to call out is that it is common for service configuration files to be called `application.yml` so that might be a little confusing to also have `application.meteor.yaml`.

ravisuhag: Yes, we can rename it to application everywhere. Since service is a concept we are already using. In terms of the name of the file, I think we can leave it to the user and the extractor can simply accept the file path.

plugins/extractors/service_yaml/README.md
sudo-suhas: Thanks for catching this!

plugins/extractors/service_yaml/build_asset.go
sudo-suhas: Not sure if that is possible. Where would we store the dataset info on the `Asset` so that the processor can access it for populating lineage? Also, I don't see it as being different from what we are doing in other extractors. We parse the data sources/sinks to construct lineage.

plugins/extractors/service_yaml/service.go
sudo-suhas: > Since service is a concept we are already using.

Could you help me understand this? Wasn't able to follow.

ravisuhag: https://github.com/odpf/proton/blob/23148b88c221d01a41aaacdff2e6700fc2d7ef25/odpf/assets/v1beta2/asset.proto#L25

sudo-suhas: https://github.com/odpf/proton/pull/214

plugins/util.go
```diff
@@ -65,7 +67,54 @@ func buildConfig(configMap map[string]interface{}, c interface{}) (err error) {
 	return err
 }
 
+func BigQueryTableFQNToURN(fqn string) (string, error) {
```
sudo-suhas, line 70: Add tests.

plugins/extractors/service_yaml/service.go
StewartJingga: > How should we handle building lineage in that case? It might be surprising if only a specific subset is supported for building lineage.

Hmm yeah that is actually true 🤔 , let's keep it as it is then for now

StewartJingga (approved):
LGTM

plugins/extractors/service_yaml/build_asset.go
StewartJingga: @ravisuhag judging from the parameter (`dd []Dataset`), it looks like this function is pretty specific to this processor.

We could generalise some of the logic so that other processors can use it but IMHO I think we can do that when the needs come.

sudo-suhas:
@StewartJingga do you want to take a final look or shall I go ahead and merge this in?

PR #428, opened 2022-11-03
feat: add script processor using Tengo

sudo-suhas:
- Implement a new 'script' processor with 'tengo' as the only supported script engine.
- Add internal packages for translating asset proto value into map[string]interface{} and for updating asset proto value with the given map.


haveiss (approved):

plugins/processors/script/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Usage
+
+```yaml
+processors:
+  - name: script
+    config:
+      engine: tengo
+      script: |
+        asset.owners = append(asset.owners || [], { name: "Big Mom", email: "big.mom@wholecakeisland.com" })
```
StewartJingga, line 20: should we also put more examples like how to merge map and loop through list? wdyt?

plugins/processors/script/tengo_script.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return p
+}
+
+func (p *Processor) Init(ctx context.Context, config plugins.Config) error {
+	if err := p.BasePlugin.Init(ctx, config); err != nil {
+		return fmt.Errorf("script processor init: %w", err)
+	}
+
+	s := tengoutil.NewSecureScript(([]byte)(p.config.Script))
```
StewartJingga, line 72: Just an idea, wdyt if we prepend the script with helper functions like below.

```
text := import("text")

merge := func(m1, m2) {
    for k, v in m2 {
        m1[k] = v
    }
    return m1
}
```

CMIIW, adding this should help users from defining their own `merge` function on every recipe.

plugins/internal/tengoutil/secure_script.go
```diff
@@ -0,0 +1,21 @@
+package tengoutil
```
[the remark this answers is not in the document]
sudo-suhas: Planning to use it for HTTP extractor as well so it will live very temporarily in `plugins/processors` .. should be okay to keep it in current location IMO

plugins/processors/script/README.md
sudo-suhas: I have added a thorough example already - https://github.com/odpf/meteor/blob/script-processor/plugins/processors/script/README.md#worked-example

Do you think this suffices or we should add a section showing script code for merging map, looping through list etc.?

plugins/processors/script/tengo_script.go
sudo-suhas: There is explicit support for defining an importable module but at this point, considering we don't have a clear idea of what all helpers could be useful, I would lean towards gathering some experience before introducing the helper module.

plugins/processors/script/README.md
StewartJingga: oh missed that part. lgtm then

plugins/processors/script/tengo_script.go
StewartJingga: Ok so the idea is, if we are going to use this `script` processor to replace our `labels` processor in all of our recipes we would need to define this `merge` everytime right?

Was thinking if we have that, users can easily migrate to `script` processor. Or we can also define `set_label` function if that is more helpful.

plugins/internal/tengoutil/secure_script.go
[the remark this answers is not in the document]
StewartJingga: I see, sure then let's keep it this way, I don't have a really strong opinion on it aside from keeping it the code close where it is being used.

plugins/processors/script/tengo_script.go
sudo-suhas: IMO, using the script processor for setting labels is overkill and it will not nearly be as clear as using the labels processor. I think it still makes sense to provide processors for common use cases that are easier to define, test and maintain from the user's perspective.

And regarding `merge`, I am not sure if we should provide that as a standard way for setting a value into the map. It _might_ be better to provide a `set_map` helper that would initialise the map if required. Might be worth analysing in a different PR.

StewartJingga (approved):
LGTM

PR #437, opened 2022-11-14
feat(bigquery): set max page size for iterators in bigquery extractor

bsushmith:


plugins/extractors/bigquery/README.md
```diff
@@ -24,62 +25,74 @@ source:
...
-| `usage_project_ids` | `[]string` | `[google-project-id, other-google-project-id]` | collecting log from defined GCP Project IDs. Default to BigQuery Project ID. | *optional* |
+| Key                      | Value      | Example                                        | Description                                                                                                                  |     |
+|:-------------------------|:-----------|:-----------------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------|:----|
+| `project_id`             | `string`   | `my-project`                                   | BigQuery Project ID                                                                                                          | *   |
+| required*                |            |                                                |                                                                                                                              |     |
```
sudo-suhas, line 37: The table formatting is broken because of a different line for required/optional. Looks like this - 
image

sudo-suhas (commented):
Changes looks good but wish we could have added tests for the changes.

plugins/extractors/bigquery/README.md
bsushmith: fixed the formatting

sudo-suhas (approved):
LGTM

sudo-suhas:
One question, is there any link between the `MaxPreviewRows` and this new config? Does it make sense to default `MaxPageSize` to 100 if our `MaxPreviewRows` is 30?

singhvikash11 (approved):

bsushmith:
> One question, is there any link between the MaxPreviewRows and this new config? Does it make sense to default MaxPageSize to 100 if our MaxPreviewRows is 30

MaxPageSize is a [hint](https://pkg.go.dev/google.golang.org/api/iterator#PageInfo) to the underlying iterator for buffering items. 

whereas MaxPreviewRows is for fetching sample rows from the bigquery table, which uses an iterator for fetching rows. so if `MaxPreviewRows` is 30, but `MaxPageSize` is 100 - then, 100 records will be fetched(buffered) and then only 30 will be saved as part of the preview in asset.

sudo-suhas:
In that case, wouldn't it be better to ensure `MaxPageSize` is less than or equal to `MaxPreviewRows`?

bsushmith:
not necessarily, as this page size is used for all iterators, and this can reduce perf for other iterators if a user sets a very low `MaxPreviewRows`. Another way would be to separate the sizes of all these iterators and expose them for user input with default values.

keeping this in the existing state, shouldn't cause any problems for preview rows.

PR #442, opened 2022-11-23
feat: add HTTP extractor

### LINKED ISSUE #413: Add HTTP Extractor

sudo-suhas:
- Add HTTP extractor, a generic extractor that uses a script capable of
  emitting any type of asset.
- Fix config error field key to support nested fields.
- Disallow import of os module from scripts.
- Remove unnecessary init method on application extractor.
- Add test helper to assert slice of protos.

Depends on #434, that needs to be merged first. 
Closes #413

### DIFF
--- a/plugins/extractors/http/execute_script.go
+++ b/plugins/extractors/http/execute_script.go
@@ -1,0 +1,169 @@
+package http
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/d5/tengo/v2"
+	"github.com/odpf/meteor/models"
+	v1beta2 "github.com/odpf/meteor/models/odpf/assets/v1beta2"
+	"github.com/odpf/meteor/plugins"
+	"github.com/odpf/meteor/plugins/internal/tengoutil"
+	"github.com/odpf/meteor/plugins/internal/tengoutil/structmap"
+	"google.golang.org/protobuf/proto"
+)
+
+func (e *Extractor) executeScript(ctx context.Context, res interface{}, emit plugins.Emit) error {
+	s := tengoutil.NewSecureScript(([]byte)(e.config.Script.Source))
+	if err := declareGlobals(s, res, emit); err != nil {
+		return fmt.Errorf("declare globals: %w", err)
+	}
+
+	c, err := s.Compile()
+	if err != nil {
+		return fmt.Errorf("compile: %w", err)
+	}
+
+	if err := c.RunContext(ctx); err != nil && !errors.Is(err, errUserExit) {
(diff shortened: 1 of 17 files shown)


plugins/extractors/http/execute_script.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func declareGlobals(s *tengo.Script, res interface{}, emit plugins.Emit) error {
+	for name, v := range map[string]interface{}{
+		"response": res,
+		"new_asset": &tengo.UserFunction{
+			Name:  "new_asset",
+			Value: newAssetWrapper(),
+		},
+		"emit": &tengo.UserFunction{
```
StewartJingga, line 42: is it possible to hide the emit context internally and when user creates the script, it would be like returning only?

sudo-suhas: Modifications done in the script do not reflect in the value in Go space. We can pull out values from the script as long as it was declared in the global scope. But when assets are declared and populated from inside a loop, it won't be possible to automatically emit it from Go.

StewartJingga:
overall LGTM, but I found a case where user might want to make another call for each item to get more metadata.

Let's say we want to fetch jobs metadata from `https://example.com/jobs`. That endpoint will give us list of jobs that we will convert into Meteor assets. Unfortunately, to get other metadata of those jobs such as lineage, we need to make another call to `https://example.com/jobs/job-id` for each job. This will not be possible right now in current state.

Is it possible to have that capability in the `http` extractor? And yes this is outside of this PR's scope, but asking it here in case there is actually a way to handle that.

@sudo-suhas

sudo-suhas:
The case you mentioned is definitely valid but we will need to think about it a bit more and implement it carefully. There could be security implications; at this point, it is not entirely clear what the scope of recursive HTTP calls and handling would be. There is no easy way to expose it currently. I would consider it outside the scope of this PR and prefer addressing it later.

StewartJingga (approved):

PR #457, opened 2023-01-04
feat(bigquery): save policy tags in column attributes

### LINKED ISSUE #456: fetch column policy tags as part of bigquery table extractor.
 Bigquery supports column annotations for enforcing column-level security constraints. It provides fine-grained access to sensitive columns using policy tags, or type-based classification.

Fetch column [policytags](https://pkg.go.dev/cloud.google.com/go/bigquery#PolicyTagList) for a given table and save it in the bigquery asset metadata for discovery.

bsushmith:
* bigquery supports column level policy tags for securing access to sensitive data in columns through policyTags. fetch the policy tags assigned to a given column and save it as attributes in asset only when the column has policy tags defined.

closes #456


plugins/extractors/bigquery/bigquery.go
```diff
@@ -477,6 +483,13 @@ func (e *Extractor) getColumnMode(col *bigquery.FieldSchema) string {
 	}
 }
 
+func (e *Extractor) getPolicyTagList(col *bigquery.FieldSchema) []string {
```
sudo-suhas, line 486: This is unnecessary. `len` will work fine with a `nil` slice as well and return 0 as expected.

bsushmith:
I had to make additional changes in this and rebased with master. It is causing issues since the asset type has changed from v1beta1 to v1beta2, which doesn't have `Properties.Tags` in Column.

plugins/extractors/bigquery/bigquery.go
sudo-suhas: Apologies, looks like this review comment was incorrect.

sudo-suhas (approved):

plugins/extractors/bigquery/bigquery.go
```diff
@@ -477,6 +502,31 @@ func (e *Extractor) getColumnMode(col *bigquery.FieldSchema) string {
...
+		}
+	}
+
+	// convert to []interface{}, so that it can be constructed into a proto
+	ptSlice := make([]interface{}, len(pt))
+	for i, s := range pt {
+		ptSlice[i] = s
+	}
+
+	return ptSlice
```
sudo-suhas, line 527:
```suggestion
	if col.PolicyTags == nil || e.policyTagClient == nil {
		return nil
	}
	
	pt := make([]interface{}, 0, len(col.PolicyTags.Names))
	for _, name := range col.PolicyTags.Names {
		policyTag, err := e.policyTagClient.GetPolicyTag(ctx, &datacatalogpb.GetPolicyTagRequest{Name: name})
		if err != nil {
			e.logger.Error("error fetching policy_tag", "policy_tag", name, "err", err)
			continue
		} 
		
		pt = append(pt, fmt.Sprintf("policy_tag:%s:%s", policyTag.DisplayName, policyTag.Name))
	}

	return pr
```

sudo-suhas (approved):

PR #469, opened 2023-01-31
feat: add GCS sink

Chief-Rishab:


plugins/sinks/gcs/client.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"cloud.google.com/go/storage"
+	"github.com/pkg/errors"
+	"google.golang.org/api/option"
+)
+
+type GCSClient interface {
```
kushsharma, line 11: We can call this `Writer`?

plugins/sinks/gcs/README.md
```diff
@@ -0,0 +1,41 @@
+# GCS
+
+Sinks json data to a file in `ndjson` format in a Google Cloud Storage bucket
```
kushsharma, line 3: Maybe we can rephrase this as:

Sinks json data to a file as [ndjson](http://ndjson.org/) format in Google Cloud Storage bucket

plugins/sinks/gcs/client.go
```diff
@@ -0,0 +1,49 @@
...
+func newGCSClient(ctx context.Context, serviceAccountJSON []byte, bucketname string, filepath string) (GCSClient, error) {
+	client, err := storage.NewClient(ctx, option.WithCredentialsJSON(serviceAccountJSON))
+	if err != nil {
+		return nil, errors.Wrap(err, "error in creating client")
+	}
+
+	writer := client.Bucket(bucketname).Object(filepath).NewWriter(ctx)
+
+	return &gcsClient{
+		client: client,
```
kushsharma, line 30: I don't see client getting used anywhere, should we get rid of it?

plugins/sinks/gcs/client.go
```diff
@@ -0,0 +1,49 @@
...
+	WriteData([]byte) error
+	Close() error
+}
+
+type gcsClient struct {
+	client *storage.Client
+	writer *storage.Writer
+}
+
+func newGCSClient(ctx context.Context, serviceAccountJSON []byte, bucketname string, filepath string) (GCSClient, error) {
```
kushsharma, line 21: You should return `*gcsClient` instead of interface.
> Remember always return struct and accept interfaces in args

plugins/sinks/gcs/client.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return &gcsClient{
+		client: client,
+		writer: writer,
+	}, nil
+}
+
+func (c *gcsClient) WriteData(jsonBytes []byte) error {
+	if _, err := c.writer.Write(jsonBytes); err != nil {
+		return errors.Wrap(err, "error in writing json data to an object")
```
kushsharma, line 37: the client should not know if the data is json or something else, its just raw bytes for it.

plugins/sinks/gcs/client.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+func (c *gcsClient) Close() error {
+	if err := c.writer.Close(); err != nil {
+		return errors.Wrap(err, "error closing the writer")
+	}
+
+	return nil
+}
```
kushsharma, line 49: Maybe simplify this as
```
func (c *gcsClient) Close() error {
	return  c.writer.Close()
}
```

plugins/sinks/gcs/client.go
Chief-Rishab: yes, its's only used to create the writer, will update the struct

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		s.config.ServiceAccountJSON = string(serviceAccountJSON)
+	}
+	return nil
+}
+
+func (s *Sink) resolveBucketandObjectNames() (string, string) {
+	dirs := strings.Split(s.config.Path, "/")
+	bucketname := dirs[0]
+	timestamp := time.Now().Format("2006.01.02 15:04:05")
```
kushsharma, line 107: I would suggest using a pre-defined time RFC format, specially the one without space.

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if s.config.ObjectPrefix != "" {
+		s.config.ObjectPrefix = s.config.ObjectPrefix + "-"
+	}
```
kushsharma, line 111: What if user has provided prefix as `hello-`, you will add double hyphens?

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+
+	objectname := s.config.ObjectPrefix + timestamp + ".ndjson"
```
kushsharma, line 113: use `fmt.Sprintf` to build strings.

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(dirs) > 1 {
+		objectname = dirs[len(dirs)-1] + "/" + s.config.ObjectPrefix + timestamp + ".ndjson"
```
kushsharma, line 115: use `fmt.Sprintf` to build strings.

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if s.client, err = newGCSClient(ctx, []byte(s.config.ServiceAccountJSON), bucketname, objectname); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *Sink) validateServiceAccountKey() error {
+
```
kushsharma, line 89: unnecessary space?

plugins/sinks/gcs/gcs.go
Chief-Rishab: It should be a single hyphen only. I will add a check for that

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return nil
+}
+
+func (s *Sink) Close() (err error) {
+	if err := s.client.Close(); err != nil {
+		return err
+	}
+	return nil
+}
```
kushsharma, line 153:
```go
return s.client.Close()
```

plugins/sinks/gcs/gcs.go
Chief-Rishab: RFC3339? Output format is "2006-01-02T15:04:05Z07:00" ?

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+		serviceAccountJSON, err := base64.StdEncoding.DecodeString(s.config.ServiceAccountBase64)
+		if err != nil || len(serviceAccountJSON) == 0 {
+			return errors.Wrap(err, "failed to decode base64 service account")
+		}
+		s.config.ServiceAccountJSON = string(serviceAccountJSON)
+	}
+	return nil
+}
+
+func (s *Sink) resolveBucketandObjectNames() (string, string) {
```
kushsharma, line 104: A better name could be `resolveBucketPath`

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+	dirs := strings.Split(s.config.Path, "/")
+	bucketname := dirs[0]
```
kushsharma, line 106: A notation pretty common in object storage is `gcs://bucketname` or `s3://bucketname` I guess this line would fail here. Can we use `url.Parse(...)`?

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+	objectprefix := s.config.ObjectPrefix 
+	
+	if objectprefix != "" && objectprefix[len(objectprefix)-1] !='-'{
+		objectprefix = fmt.Sprintf("%s-", objectprefix)
+	}
+	
+	objectname := fmt.Sprintf("%s%s.ndjson", objectprefix, timestamp)
+	
+	if len(dirs) > 1 {
+		objectname = fmt.Sprintf("%s/%s%s.ndjson", dirs[len(dirs)-1], objectprefix, timestamp)
```
kushsharma, line 118: How will `dirs[len(dirs)-1]` work for a path like `gcs://bucketname/path1/path2/`

plugins/sinks/gcs/gcs.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		if err := s.writer.WriteData([]byte("\n")); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Sink) Close() (err error) {
```
kushsharma, line 151: `(err error)` could be just `error`

kushsharma (changes requested):

plugins/sinks/gcs/gcs.go
Chief-Rishab: The default path format I assumed was the one I got from GCS bucket/folder copy path options. If that's the required case, will handle the gcs prefix

kushsharma:
LGTM. Nice work.

PR #470, opened 2023-02-01
feat: enhance HTTP extractor's script options

sudo-suhas:
- Allow configuration of the max_allocs and max_const_objects that limit memory allocation in the script env.
- Pass the recipe's scope as a script global so that it can be used to construct the asset's URN if needed.

### DIFF
--- a/plugins/extractors/http/README.md
+++ b/plugins/extractors/http/README.md
@@ -57,13 +57,15 @@ source:
 
 ## Inputs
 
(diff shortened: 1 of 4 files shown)


haveiss (approved):

sudo-suhas:
@ravisuhag would it be possible to release this change as `v0.7.2`?

ravisuhag:
@sudo-suhas yes, feel free to create a release.

Issue #462, opened 2023-01-27
capture nested fields in `RECORD` type col in bigquery extractor


bsushmith:
**Is your feature request related to a problem? Please describe.**
Meteor doesn't capture nested schema in a `RECORD` Type column in bigquery. Bigquery supports repeated and nested columns. Documentation - [link](https://cloud.google.com/bigquery/docs/nested-repeated#define_nested_and_repeated_columns)

For example: 
If the below is the DDL of the bigquery table - 
```sql
CREATE TABLE `sample_project.sample_dataset.test_with_tags` (   
    non_record_field STRING,   
    record_field STRUCT<field_1 STRING, field_2 BOOL> 
); 

```

The schema of the table once extracted through meteor looks like this - 

```json
{
  "schema": {
    "columns": [
      {
        "name": "non_record_field",
        "data_type": "STRING",
        "properties": {
          "attributes": {
            "mode": "NULLABLE"
          }
        },
        "is_nullable": true
      },
      {
        "name": "record_field",
        "data_type": "RECORD",
        "properties": {
          "attributes": {
            "mode": "NULLABLE"
          }
        },
        "is_nullable": true
      }
    ]
  }
}

```

**Describe the solution you'd like**
Meteor should also capture nested structure of the `RECORD` type column in bigquery. and resultant schema field in json can look something like this -

```json
{
  "schema": {
    "columns": [
      {
        "name": "non_record_field",
        "data_type": "STRING",
        "properties": {
          "attributes": {
            "mode": "NULLABLE"
          }
        },
        "is_nullable": true
      },
      {
        "name": "record_field",
        "data_type": "RECORD",
        "properties": {
          "attributes": {
            "mode": "NULLABLE"
          }
        },
        "is_nullable": true,
        "schema" : {
          "columns": [
            {
              "name": "field_1",
              "data_type": "STRING",
              "properties": {
                "attributes": {
                  "mode": "NULLABLE"
                }
              },
              "is_nullable": true
            },
            {
              "name": "field_2",
              "data_type": "BOOLEAN",
              "properties": {
                "attributes": {
                  "mode": "NULLABLE"
                }
              },
              "is_nullable": true
            }
          ]
        }
      }
    ]
  }
}

```

PR #476, opened 2023-03-02
feat(bigquery): add support to capture schema of nested fields

anjali9791:
Closes: https://github.com/odpf/meteor/issues/462

_edit: change `refer` to `closes`_


anjali9791:
The http_extractor_test.go failing in plugins-test here is passing on my local. Here is a screenshot for the same. 
Screenshot 2023-03-02 at 10 06 33 PM

sudo-suhas:
Could you change the `expectedErr` for that test to `Error:Field validation for 'url' failed on the 'required' tag`? The failure is because of the different ordering between `content_type` and `accept`.

plugins/extractors/bigquery/bigquery.go
```diff
@@ -366,6 +365,10 @@ func (e *Extractor) buildColumn(ctx context.Context, field *bigquery.FieldSchema
 		Attributes:  utils.TryParseMapToProto(attributesMap),
 	}
 
+	if field.Schema != nil {
```
sudo-suhas, line 368: `field.Schema` is of the following type:

```go
// Schema describes the fields in a table or query result.
type Schema []*FieldSchema
```

So check for `len(field.Schema) > 0` instead of the `nil` check.

anjali9791: Done

bsushmith (approved):

PR #478, opened 2023-03-27
feat(extractor-plugins): add crawler specifications

Chief-Rishab:
Adding option to exclude resources from being crawled from extractors
- [x] MongoDB 
- [x] CouchDB
- [x] MySQL
- [x] MsSQL
- [x] Snowflake (TODO: edit testcases)
- [x] Presto
- [x] MariaDB
- [x] GCS
- [x] Cassandra
- [x] Clickhouse
- [x] Grafana

**Note**: Metabase and Snowflake API responses have changed and the meteor extractors needs to be edited to support different versions

### DIFF
--- a/plugins/extractors/cassandra/README.md
+++ b/plugins/extractors/cassandra/README.md
@@ -10,34 +10,39 @@ source:
     password: 1234
     host: localhost
     port: 9042
+    exclude:
+      keyspaces: [mykeyspace]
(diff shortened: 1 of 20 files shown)


kushsharma (approved):
LGTM.

Minor nit: `sqlutil` method of building boolmap has nothing to do with SQL. Should we pull it out to plugins util? Just check if that causes cycle dependency.

